PR #28, opened 2015-01-06
Ref #8548 - optionally configure gutterball

dustints:


dustints:
abandoning because of new gutterball plugin in puppet-katello
gutterball plugin will be installed from katello-installer

PR #30, opened 2015-01-28
Ref #9134 - foreman-gb plugin sets configs for foreman-gutterball

dustints:
depends on https://github.com/Katello/puppet-gutterball/pull/6

### DIFF
--- a/manifests/plugin/foreman_gutterball.pp
+++ b/manifests/plugin/foreman_gutterball.pp
@@ -1,12 +1,18 @@
 # foreman_gutterball plugin
 class katello_devel::plugin::foreman_gutterball {
   $path = "${katello_devel::deployment_dir}/foreman-gutterball"
+  $foreman_settings_path = "${katello_devel::deployment_dir}/foreman/config/settings.plugins.d"
 
   vcsrepo { $path:
     ensure   => present,
     provider => git,
     source   => 'https://github.com/Katello/foreman-gutterball.git',
     user     => $katello_devel::user
+  } ~>
+  class { 'katello::plugin::gutterball::config':
+    foreman_plugins_dir => $foreman_settings_path,
+    foreman_user        => $katello::user,
+    foreman_group       => $katello::group,
   }
 
   Class[ 'katello_devel::install' ] ->


manifests/plugin/foreman_gutterball.pp
```diff
@@ -7,6 +7,11 @@
     provider => git,
     source   => 'https://github.com/Katello/foreman-gutterball.git',
     user     => $katello_devel::user
+  } ~>
+  class { 'katello::plugin::foreman_gutterball_config':
+    foreman_plugins_dir => "${path}/config/gutterball.yml",
```
dustints: @komidore64 do you think this is right for a dev install?

manifests/plugin/foreman_gutterball.pp
```diff
@@ -1,12 +1,18 @@
 # foreman_gutterball plugin
 class katello_devel::plugin::foreman_gutterball {
   $path = "${katello_devel::deployment_dir}/foreman-gutterball"
+  $foreman_settings_path = "${katello_devel::deployment_dir}/foreman/config/settings.plugins.d"
```
komidore64: this is the correct dir for a devel install, not totally sure for prod install. still need to check. would it be possible to symlink the one from the foreman-gutterball checkout to here?

```sh
mkdir -p ~/foreman/config/settings.plugins.d/
ln -s ~/foreman-gutterball/config/foreman_gutterball.yaml ~/foreman/config/settings.plugins.d/
```

manifests/plugin/foreman_gutterball.pp
```diff
@@ -1,12 +1,18 @@
...
 
   vcsrepo { $path:
     ensure   => present,
     provider => git,
     source   => 'https://github.com/Katello/foreman-gutterball.git',
     user     => $katello_devel::user
+  } ~>
+  class { 'katello::plugin::foreman_gutterball_config':
+    foreman_plugins_dir => "${foreman_settings_path}/gutterball.yml",
```
komidore64: the filename should be (afaik) the same as the name of the plugin, so we'll want `foreman_gutterball.yaml` here.

dustints: @komidore64 oh woops!...yeah not supposed to include the `gutterball.yml`

manifests/plugin/foreman_gutterball.pp
dustints: @komidore64 i think i have the right dir for prod:
https://github.com/Katello/puppet-katello/blob/afa0d0f69acfae9a7d27710301671e1a102962d4/manifests/params.pp#L47
that dir then gets copied over to /usr/share/foreman/config/settings.plugins.d/

dustints: @komidore64 wondering if we need the config file from foreman-gutterball or just keep all the configs here, in puppet land?

manifests/plugin/foreman_gutterball.pp
dustints: @komidore64  addressed

manifests/plugin/foreman_gutterball.pp
dustints: @komidore64 did the concat in https://github.com/Katello/puppet-katello/pull/58 like we talked about.

didn't do a symlink because it might get wiped on next gem version.

komidore64: alright cool

komidore64:
ack

PR #31, opened 2015-02-06
refs #9060 - configure qpid::client with params

stbenjam:


jlsherrill:
Note: requires https://github.com/Katello/puppet-qpid/pull/12

jlsherrill:
ACK

PR #32, opened 2015-02-06
fixes #9254 - Remove consumer cert generation Katello module in favor of...

jlsherrill:
... capsule


stbenjam:
APJ

(for the record this is implement this change in devel https://github.com/Katello/puppet-katello/pull/53)

PR #35, opened 2015-03-15
Registers the Capsule during development setup.

ehelms:


ehelms:
I tested this on an existing devel setup that I had and everything worked as well.

jlsherrill:
ACK, though this is still an issue http://projects.theforeman.org/issues/9702

PR #36, opened 2015-04-07
Refs #10086: Try multiple gpg sites when installing rvm.

ehelms:


templates/install_rvm.sh.erb
```diff
@@ -0,0 +1,14 @@
...
+USER=<%= @user %>
+
+SERVERS=(
+    hkp://keys.gnupg.net
+    hkp://pgp.mit.edu
+)
+
+for server in "${SERVERS[@]}"
+do
+    su -c "gpg --keyserver ${server} --recv-keys D39DC0E3" - $USER
```
stbenjam: You could also just include the GPG key itself if there's some kind of network problem that's preventing the downloading.  Up to you.  

(RVM's gpg signing is still of dubious value anyway if the script we're getting from `https://get.rvm.io` isn't signed itself. Nothing's stopping the @TheNSABackdoorCommittee from doing a MITM attack there.)

stbenjam:
APT, looks like just lint style stuff.

jlsherrill:
Tested and worked fine for me APT too

templates/install_rvm.sh.erb
ehelms: What started this was the first server being unreachable. So if the key
ever changes we'd have to update whereas this should hopefully break never
again with the redundancy.

PR #37, opened 2015-04-08
Refs #10086: Exit loop if a particular server returns successfully.

ehelms:


templates/install_rvm.sh.erb
```diff
@@ -3,12 +3,16 @@
...
     hkp://pgp.mit.edu
+    hkp://keys.gnupg.net
 )
 
 for server in "${SERVERS[@]}"
 do
-    su -c "gpg --keyserver ${server} --recv-keys D39DC0E3" - $USER
+    success=`su -c "gpg --keyserver ${server} --recv-keys D39DC0E3" - $USER`
+
+    if ($success); then
```
ekohl: are the braces needed here?

ekohl:
Minor nitpick, but :+1:

ehelms:
@ekohl thanks for the review, updated
@stbenjam or @jlsherrill  either of you good with this slight update to add to the installer PR?

templates/install_rvm.sh.erb
```diff
@@ -3,12 +3,16 @@
...
 )
 
 for server in "${SERVERS[@]}"
 do
-    su -c "gpg --keyserver ${server} --recv-keys D39DC0E3" - $USER
+    success=`su -c "gpg --keyserver ${server} --recv-keys D39DC0E3" - $USER`
+
+    if $success; then
+        break
+    fi
```
jlsherrill: would it make more sense to check the return value of the command?  ($?)

Testing locally i still hit break even if success is empty (implying no output)

jlsherrill:
APJ

PR #46, opened 2015-10-08
Fixes #12110 - updating puppet-pulp to use is_parent => true param

johnpmitsch:


stbenjam:
Thanks @johnpmitsch !

PR #48, opened 2015-11-17
Refs #12475: Configure AMQP for Candlepin since it is off by default

ehelms:


thomasmckay:
ACK! Events being received again.

PR #50, opened 2015-11-19
Fixes #10291 - removes elasticsearch

cfouant:


ehelms:
ACK, thanks @cfouant !

PR #53, opened 2016-01-29
Rely on foreman_proxy instead of capsule module

ehelms:


ehelms:
This needs updating after removing the puppet-foreman_proxy from puppet-capsule.

cfouant:
ACK

PR #56, opened 2016-02-05
Refs #13189 - generate pulp client certs

jlsherrill:


ehelms:
We no longer set those ssl setting on the Pulp class in puppet-katello so you could remove them here

jlsherrill:
@ehelms updated

ehelms:
ACK

PR #57, opened 2016-02-09
Refs #13607 - Removed pulp.conf

parthaa:
The template is now removed and centralized in puppet-pulp

Check https://github.com/Katello/puppet-pulp/pull/124


ekohl:
You can also remove the ```$apache_version``` from config.pp

ekohl:
And while you're at it, I think you can also remove the include above it, but do double check that.

parthaa:
> You can also remove the $apache_version from config.pp

good catch. Will do thanks!

manifests/init.pp
```diff
@@ -140,6 +140,7 @@
     enable_puppet          => true,
     enable_docker          => true,
     enable_parent_node     => true,
+    ca_cert                => $::certs::ca_cert,
```
ehelms: This doesn't need tobe set -- see https://github.com/Katello/puppet-katello/pull/102

manifests/config.pp
```diff
@@ -1,10 +1,5 @@
 # Configuration for Katello development
 class katello_devel::config {
-
-  # Required until the katello.yml.erb changes to '@variable' syntax
-  include ::katello::params
```
ehelms: I think we still need this as indicated by the comment

ehelms: I take that back, @ekohl is right we updated katello.yaml properly so this is good to go

ehelms:
ACK - thanks @parthaa !

PR #61, opened 2016-02-13
Bump rvm base version to 2.2.4

ehelms:


beav:
@ehelms the change itself looks good. Can you update the commit message to add why the bump is needed?

ehelms:
@beav updated

beav:
ACK

PR #64, opened 2016-02-18
fixes #13451 - Enables lazy sync

cfouant:


manifests/init.pp
```diff
@@ -137,6 +137,7 @@
     consumers_crl          => $candlepin::crl_file,
     manage_broker          => false,
     manage_httpd           => false,
+    manage_httpd           => true,
```
ekohl: This doesn't look right.

daviddavis:
ACK

PR #69, opened 2016-03-23
Set enable_parent_node => false to match production

akofink:


ehelms:
ACK - thanks @akofink !

PR #70, opened 2016-03-24
Add pry-byebug and puma in a Gemfile.local.rb for foreman

akofink:


thomasmckay:
Will these conflict with other debug gems? I use ruby-debug-ide and I have found that pry conflicts when included.

stbenjam:
I'd set `:replace => false` so existing Gemfile.local.rb's don't get overwritten, and if a dev prefers a different debugger (like @thomasmckay's comment), then they can pick a new one without risking the installer wiping it out again later.

ehelms:
We could name it something else since anything will be picked up in
bundler.d similar to plugin files.

templates/Gemfile.local.rb.erb
```diff
@@ -0,0 +1,4 @@
+group :development do
```
ehelms: Will this clash with the existing development group?

templates/Gemfile.local.rb.erb
```diff
@@ -0,0 +1,4 @@
...
+  gem 'pry-byebug'
+  gem 'puma'
```
ehelms: This could be puma-rails instead.

templates/Gemfile.local.rb.erb
```diff
@@ -0,0 +1,4 @@
+group :development do
+  gem 'pry-byebug'
```
ehelms: Does this install byebug ? I know some devs use that directly instead of pry

akofink:
Maybe we could use local environment variables to enable or disable groups using Bundler's new [`install_if`](https://github.com/bundler/bundler/blob/6342a6bf2bb4a51902fdea14ef6d80c1fb6847e0/spec/install/gems/install_if.rb#L8) feature. Is there a way to pass environment variables to the Vagrant box?

ehelms:
Ehh, I'm OK with some sane defaults and letting developers customize there after.

templates/Gemfile.local.rb.erb
daviddavis: [Yep](https://github.com/deivid-rodriguez/pry-byebug/blob/master/pry-byebug.gemspec#L23).

templates/Gemfile.local.rb.erb
daviddavis: Nope.

templates/Gemfile.local.rb.erb
daviddavis: I'm a bit hesitant to force everyone to use puma given. Unlike byebug which is pretty much the only gem for debugging ruby 2.0+, there are many webservers for rails.

daviddavis:
@thomasmckay how are you not including pry given that [it's in foreman's Gemfiles](https://github.com/theforeman/foreman/blob/develop/bundler.d/development.rb#L11)?

akofink:
Ok, it seems like this isn't warranted.

ehelms:
I think this just needs a name change from Gemfile.local.rb and it's good to go. I don't think anyone said this isn't wanted.

stbenjam:
> I think this just needs a name change from Gemfile.local.rb and it's good to go.

Where would you add it then? If you put it in something like debugger.local.rb or whatever, it'll just get added every time the installer runs.  I like the idea of Gemfile.local.rb (but :replace => false as mentioned in https://github.com/Katello/puppet-katello_devel/pull/70#issuecomment-200598013), so you get a good default, but can change it without worrying about it coming back.

stbenjam:
And definitely a good change IMHO, dev install being more useful out of the box is a good thing.

thomasmckay:
ACK from me. It is pry-debugger that conflicts in my dev environment, not pry.

ehelms:
@stbenjam that's fair enough to start, the downside is that you cannot "pass along" changes via the installer. In other words, "we've updated the standard devel environment please update installer and run". And while it does not do that that well today, it has been a goal of mine. I am also open to a strategy of "reinstall" your dev environment vs "live updating" of it.

stbenjam:
`rm Gemfile.local.rb` and re-run the installer would get you the new Gemfile, but I see your point.  I don't particularly have any strong opinions about it, some other .local.rb is fine too.

akofink:
Ultimately, it isn't that difficult to add gems and run `bundle install` once the vagrant box is up. This kind of change would be more useful if we destroyed boxes more often, but it seems like most people get their box up and keep it up if possible.

ehelms:
Yes and no. Why do people keep their boxes up and running for so long? Traditionally this has been because setting up a new environment takes time. However, with a puppet based devel installer, even if users keep the same box longer, we should be able to disseminate updates through the installer so that existing environments are latest/greatest. If you are reluctant to re-open, then I will do it.

akofink:
Ok, I'll reopen, but I think we should open discussion about which dev gems we should include by default and why. I know that I use pry for debugging and puma makes the dev server quite a bit more responsive, but anything but Webrick would probably be just as good.

akofink:
Closing this PR. If we want pry/puma in the future, we can reopen, but it's not that important to me.

PR #71, opened 2016-03-30
Refactor to a plugin manifest

ehelms:


ehelms:
My eventual goal is to support specifying all the major plugins with flags to enable or disable their inclusion so that a devel install encapsulates a wider range of possibilities for development.

manifests/git_repo.pp
```diff
@@ -0,0 +1,22 @@
...
+  $source,
+  $github_username = undef,
+) {
+
+  validate_string($source)
+
+  $sources = {origin => "https://github.com/${source}.git"}
+
+  if $github_username != undef {
+    $sources[$github_username] = "https://github.com/${github_username}/${title}.git"
```
ehelms: This is probably up for debate as to whats the default for user forks: ssh or https

stbenjam: I prefer SSH, can it be configurable?

manifests/git_repo.pp
```diff
@@ -0,0 +1,22 @@
+# Setup and configure a git repo
+define katello_devel::git_repo(
+  $source,
+  $github_username = undef,
+) {
+
+  validate_string($source)
+
+  $sources = {origin => "https://github.com/${source}.git"}
```
stbenjam: Do most devs do it this way? 

My origin is usually my fork and the upstream remote is called `remote`.

manifests/init.pp
```diff
@@ -36,6 +36,8 @@
 # $enable_ostree::        Boolean to enable ostree plugin. This requires existence of an ostree install.
 #                         type:boolean
 #
+# $github_username::      Github username to add remotes for
+#
```
stbenjam: Could we enable arbitrary pluigns? Like a param here that takes an array of plugins to enable? Then you can do:

`katello_devel::plugin { $enable_plugins:}` and it'll create them all.  Not sure how this would work with the inconsistencies with foreman plugins, foreman_docker gem is in a foreman-docker repo.

manifests/git_repo.pp
ekohl: I'm in a similar position, but I use my username instead of origin. ```hub fork``` sets it up in the same way.

manifests/git_repo.pp
ekohl: You can always add the following to your ```~/.gitconfig```:
```
[url "git@github.com:"]
	pushInsteadOf = "https://github.com/"
```

Or if you want to be more specific you can add katello after it.

manifests/git_repo.pp
ehelms: We can push for convention through this :) If a more common idiom is to have username and upstream then I am all for that. Some retraining will have to occur but that's fine in my opinion.

manifests/init.pp
ehelms: I see the benefit of thought, but as you point out the dash vs hyphen downside. That will also require people to think and know what plugins they want as opposed to just seeing options for them. Which could be OK, I just tend towards the obvious but am happy to be steered towards the flexible.

manifests/git_repo.pp
stbenjam: Can it be customized? I think most people (I do anyway) prefer their particular git habits and don't want to be forced into a way of doing things.

manifests/init.pp
stbenjam: I was thinking about this the other day, if people give the git repo name (foreman-docker) to puppet, just replace the - with a _ in the gem name since *those* are consistent.

I get the point about enabling particular plugins but it's a lot to maintain especially when stuff is popping up all the time.

ehelms: Granted, we could just have both -- the curated set and then the ability to do arbitrary.

manifests/git_repo.pp
ehelms: Anything is possible, we just need a base pattern we want to go with and then let those who wish to customize there after.

manifests/init.pp
```diff
@@ -36,6 +36,15 @@
...
+# $use_ssh_fork::         If true, will use SSH to configure Github fork, otherwise HTTPS.
+#                         type:boolean
+#
+# $fork_remote_name::     Name of the remote that represents your fork
+#
+# $upstream_remote_name:: Name of the remove that represents the upstream repository
+#
```
ehelms: Added params here for the cases requested @stbenjam and @ekohl

stbenjam: Thanks! :+1: :+1: :+1:

manifests/git_repo.pp
```diff
@@ -0,0 +1,28 @@
+# Setup and configure a git repo
+define katello_devel::git_repo(
+  $source,
+  $github_username = undef,
+) {
+
+  validate_string($source)
+
+  $sources = {"${$katello_devel::upstream_remote_name}" => "https://github.com/${source}.git"}
```
ekohl: The array key here looks weird to me. Why is the string needed? Maybe this should also be a parameter to this define (and default to ```$katello_devel::upstream_remote_name```)?

ehelms: Done

manifests/config.pp
```diff
@@ -21,12 +14,12 @@
...
-    ensure  => file,
-    content => template('katello/katello.yaml.erb'),
-    owner   => $katello_devel::user,
-    group   => $katello_devel::group,
-    mode    => '0644',
   }
+
+  katello_devel::plugin { 'katello/katello': }
+
+  $extra_plugins.each |String $plugin| {
```
ekohl: Isn't this future parser only? I think this works exactly the same in older puppet versions:
```puppet
katello_devel::plugin { $extra_plugins: }
```

manifests/git_repo.pp
```diff
@@ -0,0 +1,29 @@
+# Setup and configure a git repo
+define katello_devel::git_repo(
+  $source,
+  $github_username = undef,
+  $upstream_remote_name = $katello_devel::upstream_remote_name
+) {
+
+  validate_string($source)
+
+  $sources = {upstream_remote_name => "https://github.com/${source}.git"}
```
ekohl: I think you forgot a ```$``` here.

manifests/config.pp
ehelms: Yea, I misread the docs and push to test it locally. I also tested your suggestion and that seemed to not do what I expected :/

```
[ERROR 2016-05-06 10:35:27 verbose]  Could not find template 'katello/.yaml.erb' at /usr/share/katello-installer-base/modules/katello_devel/manifests/plugin.pp:10 on node devel.example.com
```

Failing at this line https://github.com/Katello/puppet-katello_devel/pull/71/files/9922a3b9649bf35750a5d07e4f89bff98b01b702#diff-27f9b7e25f91571b5a1d6dcda89e134cR4

manifests/init.pp
```diff
@@ -59,10 +70,29 @@
...
   ) inherits katello_devel::params {
+
   validate_bool($enable_ostree)
+  validate_string($upstream_remote_name)
+
+  if $github_username != undef {
+    validate_bool($use_ssh_fork)
+  }
+
+  $fork_remote_name_real = $fork_remote_name ? {
```
ekohl: Wouldn't ```pick``` from stdlib do the same?

```puppet
$fork_remote_name_real = pick($fork_remote_name, $github_username)
```

manifests/config.pp
ekohl: I thought that would work. Maybe if you use ```$name``` instead of ```$title``` in ```plugin.pp```?

ehelms: Bah, it does work! Flaw in my assumptions if you look at the line I linked, that is trying to deploy the plugins settings file and it is hard coded to katello still. Which, @stbenjam makes generic plugins slightly trickier if they require configuration like Katello currently does.

stbenjam: Probably not many foreman plugins do, I only know of a couple.  Having the git checkouts with remotes right and gems loaded is a pretty good first step for now.

ekohl: Agreed, it's a good start we can iterate on if we need to.

ehelms:
Updated all comments as far as I know, included the settings for a plugin via a parameter to specify a template since Katello will need that piece of functionality. I have tested this on my existing dev server a bit but would benefit from another party testing it.

manifests/git_repo.pp
```diff
@@ -0,0 +1,29 @@
+# Setup and configure a git repo
+define katello_devel::git_repo(
+  $source,
+  $github_username = undef,
+  $upstream_remote_name = $katello_devel::upstream_remote_name
```
ekohl: nit: trailing comma

manifests/git_repo.pp
```diff
@@ -0,0 +1,29 @@
...
+) {
+
+  validate_string($source)
+
+  $sources = {"${upstream_remote_name}" => "https://github.com/${source}.git"}
```
ekohl: Doesn't ```{$upstream_remote_name => "https://github.com/${source}.git"}``` work?

manifests/init.pp
```diff
@@ -61,11 +72,29 @@
...
+
   validate_bool($enable_ostree)
+  validate_string($upstream_remote_name)
+
+  if $github_username != undef {
+    validate_bool($use_ssh_fork)
+  }
+
+  
+  $fork_remote_name_real = $pick($fork_remote_name, $github_username)
```
ekohl: I think you mean ```pick``` instead of ```$pick```.

manifests/init.pp
```diff
@@ -61,11 +72,29 @@
...
+
   $enable_ostree = $katello::params::enable_ostree,
   $candlepin_event_queue = $katello_devel::params::candlepin_event_queue,
 
+  $github_username = $katello_devel::params::github_username,
+  $use_ssh_fork = $katello_devel::params::use_ssh_fork,
+  $fork_remote_name = $katello_devel::params::fork_remote_name,
+  $upstream_remote_name = $katello_devel::params::upstream_remote_name,
+ 
+  $extra_plugins = $katello_devel::params::extra_plugins
```
ekohl: trailing comma

manifests/plugin.pp
```diff
@@ -0,0 +1,33 @@
...
+) {
+
+  $split_array = split($name, '/')
+  $github_organization = $split_array[0]
+  $plugin = $split_array[1]
+
+  if $settings_template {
+    file { "${katello_devel::deployment_dir}/foreman/config/settings.plugins.d/${plugin}.yaml":
+      ensure   => file,
+      content  => template($settings_template),
```
ekohl: Does ```template(undef)``` work?

ehelms:
Updated, thanks @ekohl for pointing out my silly mistakes :)

manifests/init.pp
```diff
@@ -38,6 +38,17 @@
...
+# $github_username::      Github username to add remotes for
+#
+# $use_ssh_fork::         If true, will use SSH to configure Github fork, otherwise HTTPS.
+#                         type:boolean
+#
+# $fork_remote_name::     Name of the remote that represents your fork
+#
+# $upstream_remote_name:: Name of the remove that represents the upstream repository
+#
+# $extra_plugins::        Array of Github namespace/repo plugins to setup and configure from git
```
ekohl: type:array

ekohl: and validate_array

ekohl:
Mostly :+1:

manifests/git_repo.pp
ehelms: Throws a syntax error if not wrapped in quotes.

ehelms:
Updated to fix tests and lint and rebase with master

manifests/init.pp
ehelms: Addressed

manifests/init.pp
```diff
@@ -65,14 +77,32 @@
...
+  $fork_remote_name = $katello_devel::params::fork_remote_name,
+  $upstream_remote_name = $katello_devel::params::upstream_remote_name,
+
+  $extra_plugins = $katello_devel::params::extra_plugins,
+  
   ) inherits katello_devel::params {
+
   validate_bool($enable_ostree)
+  validate_string($upstream_remote_name)
+  validate_array(extra_plugins)
```
ekohl: You're missing a ```$``` here that's causing the test to fail.

ehelms:
@ekohl tests finally passing, if you and @stbenjam don't mind making a final pass

ekohl:
I haven't had a change to test it, but :+1:

stbenjam:
ACK from me too

PR #72, opened 2016-04-12
Specify the candlepin event queue in katello_devel params

stbenjam:
We only call katello::qpid directly, and the katello top-level namespace
isn't loaded when katello_devel is running, so the queue name
doesn't resolve.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -4,37 +4,39 @@
 #
 # === Parameters:
 #
-# $user::                 The Katello system user name
+# $user::                  The Katello system user name
 #
-# $deployment_dir::       Location to deploy Katello to in development
+# $deployment_dir::        Location to deploy Katello to in development
 #
-# $oauth_key::            The oauth key for talking to the candlepin API;
-#                         default 'katello'
+# $oauth_key::             The oauth key for talking to the candlepin API;
+#                          default 'katello'
 #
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -60,6 +60,7 @@
   $initial_location = $katello_devel::params::initial_location,
   $admin_password = $katello_devel::params::admin_password,
   $enable_ostree = $katello::params::enable_ostree,
+  $candlepin_event_queue = $katello_devel::params::candlepin_event_queue,
```
stbenjam: The reason this works is `$::katello_devel::params` inherits from `$::katello::params`.

ekohl:
ACK this makes sense to me though I didn't test it.

beav:
I get this error:

```
[FATAL 2016-04-12 17:11:15 fatal] Unable to continue because of: undocumented parameters in katello_devel: candlepin_event_queue
```

beav:
new patch works for me, clean run for devel setup!

ACK

ekohl:
There's a puppet-lint check for that. We use it in all theforeman modules so travis catches those for us.

stbenjam:
@ekohl Oh I see, we have the bits in our modulesync but it's not configured in the .sync.yml's, thanks! Filed http://projects.theforeman.org/issues/14599

stbenjam:
Thanks @beav and @ekohl, merging

PR #75, opened 2016-05-16
refs #15058 - enable mongo auth

stbenjam:


ekohl:
ACK

manifests/init.pp
```diff
@@ -64,6 +68,9 @@
   $enable_ostree = $katello::params::enable_ostree,
   $candlepin_event_queue = $katello_devel::params::candlepin_event_queue,
 
+  $pulp_db_username = $katello_devel::params::pulp_db_username,
+  $pulp_db_password = $katello_devel::params::pulp_db_password,
```
ehelms: I don't see these defined anywhere in params.pp

ekohl: They are inherited from katello::params once https://github.com/Katello/puppet-katello/pull/126 is merged.

ehelms: How? This says `katello_devel::params`

ekohl: Which in turn inherits from ```katello::params``` (https://github.com/Katello/puppet-katello_devel/blob/4d38cf0cad1515bf29fd9b5e0047094404082066/manifests/params.pp#L2). Ugly, but it works ;)

ehelms: Touche :)

stbenjam: I can explicitly add them here if you want, not a big fan of this many layers deep inheritance to figure out where the variable comes from.

ekohl: I think we have the same pattern for some other parameters as well so I'd either add them all (and remove the inheritance) or leave it as it is now.

ekohl:
ACK

PR #77, opened 2016-05-24
Merge puppet-hammer_devel into puppet-katello_devel

ehelms:
Opening this up for review and comment (needs testing still) based on the general idea and approach. This is taken from adapting https://github.com/Katello/puppet-hammer_devel and combining it with the plugin style approach declaration.

### DIFF
--- a/manifests/hammer.pp
+++ b/manifests/hammer.pp
@@ -1,0 +1,18 @@
+# == Class: puppet-hammer_devel
+#
+# Install and configure hammer-cli-katello for development
+#
+class katello_devel::hammer () {
+
+  class { 'katello_devel::hammer::config': } ~>
+  katello_devel::hammer::plugin { 'theforeman/hammer-cli-foreman':
+    config_content => template('katello_devel/hammer/foreman.yml'),
+  } ~>
+  katello_devel::hammer::plugin { 'katello/hammer-cli-katello':
+    config_content => template('katello_devel/hammer/katello.yml'),
+  } ~>
+  katello_devel::hammer::plugin { 'theforeman/hammer-cli-foreman-tasks':
+    config_content => template('katello_devel/hammer/katello.yml'),
+  }
+
+}
--- a/manifests/hammer/plugin.pp
+++ b/manifests/hammer/plugin.pp
@@ -1,0 +1,30 @@
+# Configuration for hammer development
+class katello_devel::hammer::plugin (
+  $config_content = undef,
+) {
+
+  $split_array = split($name, '/')
+  $github_organization = $split_array[0]
+  $plugin = $split_array[1]
+  $gem_name = regsubst($plugin, '-', '_')
+
+  if $config_content {
+    file { "${katello_devel::deployment_dir}/.hammer/cli.modules.d/${plugin}.yml":
+      ensure  => file,
+      content => $config_content,
+      owner   => $katello_devel::user,
+      mode    => '0644'
+    }
+  }
+
+  git_repo { $plugin:
+    source          => $title,
+    github_username => $katello_devel::github_username,
+  }
+
+  file_line { $katello_devel::hammer_gemfile:
+    line => "gem '${gem_name}', :path => '../${plugin}'",
+    path => "${katello_devel::deployment_dir}/hammer-cli/Gemfile.local"
+  }
+
+}
--- a/manifests/hammer/config.pp
+++ b/manifests/hammer/config.pp
@@ -1,0 +1,27 @@
+# Configuration for hammer development
+class katello_devel::hammer::config {
+
+  file { ["${katello_devel::deployment_dir}/.hammer/",
+          "${katello_devel::deployment_dir}/.hammer/cli.modules.d/"]:
+    ensure  => 'directory',
+    owner   => $katello_devel::user,
+    mode    => '0644'
+  }
+
+  katello_devel::hammer::plugin { 'theforeman/hammer-cli': }
+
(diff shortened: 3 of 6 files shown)


manifests/hammer.pp
```diff
@@ -0,0 +1,23 @@
+# == Class: puppet-hammer_devel
+#
+# Install and configure hammer-cli-katello for development
+#
+# === Parameters:
```
ekohl: These parameters don't actually exist.

manifests/hammer/plugin.pp
```diff
@@ -0,0 +1,24 @@
+# Configuration for hammer development
+class katello_devel::hammer::plugin (
+  $config_template = undef,
```
ekohl: For scoping I think it'd make more sense to actually pass in the content rather than the template.

manifests/hammer.pp
```diff
@@ -0,0 +1,23 @@
...
+#
+# $deployment_dir::   The directory in which to deploy hammer-cli-katello and
+#                     its dependencies
+#
+class katello_devel::hammer () {
+
+  class { 'katello_devel::hammer::config': } ~>
+  katello_devel::hammer::plugin { 'theforeman/hammer_cli': } ~>
+  katello_devel::hammer::plugin { 'theforeman/hammer_cli_foreman':
+    config_tempalte => 'katello_devel/hammer/foreman.yml',
```
ekohl: typo in template

manifests/hammer.pp
```diff
@@ -0,0 +1,23 @@
...
+  } ~>
+  katello_devel::hammer::plugin { 'katello/katello':
+    config_tempalte => 'katello_devel/hammer/katello.yml',
```
ekohl: typo in template

ekohl:
I like the general idea.

ehelms:
@daviddavis if you don't mind also taking a look since this is borrowing from your work

daviddavis:
:+1: from me. Thanks for doing this.

daviddavis:
Actually a few minor things:

1. It would be nice to setup the hammer dependencies locally although this may work without it. More info: https://github.com/Katello/hammer-cli-katello#steps
2. It would be nice to create and set the permissions on `/var/log/hammer`
3. We could maybe `bundle install` inside hammer-cli-katello

Let me know if you need help with any of those.

ehelms:
@daviddavis where should the Gemfile.local go? And when we add more hammer plugins, how do we add them? Just looking for some general thoughts.

daviddavis:
> where should the Gemfile.local go?

[It goes inside hammer-cli-katello](https://git.io/vr6OC).

> And when we add more hammer plugins, how do we add them?

You treat them the same as hammer-cli-foreman. Check it out in the deployment directory (i.e. `~/`):

https://git.io/vr6Ow

Add it to Gemfile.local:

https://github.com/Katello/puppet-hammer_devel/blob/master/templates/Gemfile.local#L3

And then optionally create a file in `.hammer/cli.modules.d`. 

This of course assumes you want to use the plugin from git. In theory you could just use the plugin as a gem by adding it to the hammer-cli-katello Gemfile.

ehelms:
I'm thinking then the Gemfile should probably go in hammer_cli since its the main module and update it each time with any plugins are that are installed via git?

daviddavis:
@ehelms I think that might work. I assume by Gemfile you mean Gemfile.local? I can try that out when I get back from vacation unless @komidore64 knows for sure.

daviddavis:
I just tested it out and it works. I usually run hammer from the hammer-cli-katello directory so I'll just have to make sure I run it from hammer-cli instead from now on.

komidore64:
it shouldn't matter where ``Gemfile.local`` is as long as the current working directory's ``Gemfile`` knows how to find it.

ehelms:
@daviddavis made some updates based on the discussion, let me know if that meets your expectations

manifests/hammer/plugin.pp
```diff
@@ -0,0 +1,30 @@
...
+class katello_devel::hammer::plugin (
+  $config_content = undef,
+) {
+
+  $split_array = split($name, '/')
+  $github_organization = $split_array[0]
+  $plugin = $split_array[1]
+  $gem_name = regsubst($plugin, '-', '_')
+
+  if $config_template {
```
daviddavis: Where is this coming from?

manifests/hammer.pp
```diff
@@ -0,0 +1,18 @@
...
+
+  class { 'katello_devel::hammer::config': } ~>
+  katello_devel::hammer::plugin { 'theforeman/hammer-cli-foreman':
+    config_content => template('katello_devel/hammer/foreman.yml'),
+  } ~>
+  katello_devel::hammer::plugin { 'katello/hammer-cli-katello':
+    config_content => template('katello_devel/hammer/katello.yml'),
+  } ~>
+  katello_devel::hammer::plugin { 'theforeman/hammer-cli-foreman-tasks':
+    config_content => template('katello_devel/hammer/katello.yml'),
```
daviddavis: Is it possible to set no `config_content` at all and just have it skip creating a yml file.

manifests/hammer/plugin.pp
ehelms: A re-factoring bug, will fix

manifests/hammer.pp
ehelms: See https://github.com/Katello/puppet-katello_devel/pull/77#discussion-diff-65393111R11

ehelms:
Closing in light of -- https://github.com/Katello/forklift/pull/259

PR #79, opened 2016-06-07
refs #15326 - revert mongo auth

stbenjam:


akofink:
Pretty straightforward change - ACK

PR #80, opened 2016-06-21
refs #15217 - puppet 4 support

stbenjam:


stbenjam:
Retriggering travis, for some reason it never ran

stbenjam:
This is good to go

manifests/git_repo.pp
```diff
@@ -7,7 +7,7 @@
 
   validate_string($source)
 
-  $sources = {"${upstream_remote_name}" => "https://github.com/${source}.git"}
+  $sources = hash([$upstream_remote_name, "https://github.com/${source}.git"])
```
ekohl: I'm surprised this is needed. Maybe it's nicer to just define ```$sources``` where you define ```$sources_real```?

stbenjam: It generated an error on puppet 4 otherwise:

```
Could not parse for environment *root*: Illegal attempt to assign via [index/key]. Not an assignable reference
```

But your suggestion looks better I think, updated.

ekohl:
Build failed on a json dependency on older rubies.

stbenjam:
Now is good to go :-)

stbenjam:
Need a small change - add katello.yaml.erb to this module as its vars are no longer in scope

ehelms:
ACK from testing perspective and fixes broken installer

stbenjam:
Awesome, thanks!

PR #82, opened 2016-08-01
Redirect http requests to https similar to production

jmontleon:


jmontleon:
I'm working to enable node registration via the OSP Director heat templates. The way they determine which version of Satellite they are registering to is via the api on the non-ssl port, so I need it available in order to expect this to work.

ehelms:
Is the expectation that you'll need to do this in production as well? This being exposing the API over non-ssl?

jmontleon:
In production mode it looks like it is fine as is. 

This is a RHEL OSP script that I'm looking at and the command they run and response is below. All they're looking for is the 200 OK in the output. I'm actually suspicious looking now that in production it would work with the ssl port even though the documentation states it should be the non-ssl port because they pass -k to curl everywhere I am looking.

That said /pub is not available on the SSL port in devel, since it's redirected to rails on port 3000, so it wouldn't be able to download the katello-ca-consumer-latest rpm. The non-ssl port I think it ends up being a 403 when it tries to look for the api without the ProxyPass. 

And they are using the same base parameter to build the api and rpm url's.

So, as is there should be no problem in production, and this small change has led to success in dev.

ping_api=http://192.168.252.1/katello/api/ping curl -L -k -s -D - -o /dev/null $ping_api
HTTP/1.1 301 Moved Permanently
Date: Mon, 01 Aug 2016 17:26:59 GMT
Server: Apache/2.4.6 (Red Hat Enterprise Linux)
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
Apipie-Checksum: 47c949e9e89543caab9b14331c092ad3
Cache-Control: no-cache
X-Request-Id: 8d7838e0-3cd4-4466-8617-3bc3bfc0f987
X-Runtime: 0.008742
X-Powered-By: Phusion Passenger 4.0.18
Location: https://192.168.252.1/katello/api/ping
Status: 301 Moved Permanently
Connection: close
Transfer-Encoding: chunked
Content-Type: text/html; charset=utf-8

HTTP/1.1 200 OK
Date: Mon, 01 Aug 2016 17:26:59 GMT
Server: Apache/2.4.6 (Red Hat Enterprise Linux)
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
Foreman_version: 1.11.0.49
Foreman_api_version: 2
Apipie-Checksum: 47c949e9e89543caab9b14331c092ad3
Cache-Control: max-age=0, private, must-revalidate
X-Request-Id: 13133331-988d-4992-be89-21c923c7f9f0
X-Runtime: 1.093239
X-Powered-By: Phusion Passenger 4.0.18
Set-Cookie: _session_id=c1fbba68a67372b1f0ce66569464e648; path=/; secure; HttpOnly
ETag: "c6077dcba90136860f2f5226a37c0443"
Status: 200 OK
Vary: Accept-Encoding
Transfer-Encoding: chunked
Content-Type: application/json; charset=utf-8

ehelms:
@jmontleon I'm more worried that by making an exception in development environments instead of trying to mimic production as best as possible we introduce further issues. By either not remembering the future why we opened this up, or by relying on a expectation that isn't met and things then break in production. 

What is the error when it tries to hit the devel API?

ekohl:
I expect it works in production because it uses the ```-L``` (```--location```) option and in production HTTP will redirect to HTTPS.

ehelms:
Makes sense, so we'd just need to add redirection to HTTPs to our development deployment. @jmontleon sound reasonable to go that route?

jmontleon:
Yes, that sounds reasonable. If it hasn't been done already I can test and submit a PR

jmontleon:
@ehelms testing an installation with the udpated line yields:
Redirect permanent /katello/api https://sat62devg.example.com/katello/api
This similarly gives me a 301 moved permanently followed by 200 OK and returns api output as expected.

templates/_http.conf.erb
```diff
@@ -1 +1,2 @@
 ProxyPass /unattended http://localhost:3000/unattended
+Redirect permanent <%= scope['katello_devel::deployment_url'] %>/api https://<%= scope['katello_devel::fqdn'] %><%= scope['katello_devel::deployment_url'] %>/api
```
ehelms: Why do we want to limit this strictly to the katello API paths? Versus just doing the whole application?

jmontleon: As far as I can tell we don't need to. I updated the redirect to redirect / to https://fqdn/

ehelms:
@jmontleon one last thought, did you test whether this would handle HTTP published repositories properly? We wouldn't want anything at `/pulp/repos` to redirect to HTTPs. Sorry that should have occurred to me earlier.

jmontleon:
@ehelms Good catch. I tried unattended and the ProxyPass rule above appeared to take precedence, but I did not consider pulp or pub, which this does affect.

I can use a rewrite to get the same effect:
RewriteEngine On
RewriteCond %{REQUEST_URI} !^\/pulp\/.*
RewriteCond %{REQUEST_URI} !^\/pulp$
RewriteCond %{REQUEST_URI} !^\/pub\/.*
RewriteCond %{REQUEST_URI} !^\/pub$
RewriteCond %{REQUEST_URI} !^\/unattended\/.*
RewriteCond %{REQUEST_URI} !^\/unattended$
RewriteCond %{HTTPS} off
RewriteRule (.*) https://sat62devg.example.com/$1 [R,L]

Are there others besides pulp, pub, and unattended to your knowledge that need to be accessible via http?

ehelms:
@jmontleon that should be the full list

jmontleon:
@ehelms I've updated the PR again. I've done a RHEV QCI deployment with these rules in place which should be pulling content from the pulp repos, kickstarts from unattended, and pulling the katello ca  rpm from pub, so I think all the exceptions have been tested. I've also done openstack deployments with node registration which would require the redirect, so I'm pretty confident this is working. Let me know if there is anything else I should look into or any additional testing I can do.

ehelms:
Looks good! Thanks @jmontleon ! We'll merge it up and fix any issues that arise after the fact (if they do).

PR #83, opened 2016-08-03
puppet4 update for `install_rvm.sh`

beav:
Previously, this would not fill in `$USER` and would end up executing rvm as
root, causing breakages and confusion.

There was a secondary bug where in some cases, `rvm_path` was being set.
However, I think its OK to assume that `install_rvm.sh` is run on a pristine
system, so I left the bugfix out since it may cause confusion if someone really
does want to set `rvm_path`.


stbenjam:
APT

PR #84, opened 2016-08-04
add `scope[]` to settings.yml for puppet4

beav:
This was being referenced improperly for puppet4, resulting in foreman-proxy
not knowing its oauth key/secret in dev installs.


stbenjam:
APT

PR #92, opened 2016-11-01
Add username in git url

sean797:
Small thing but there no need to type your username in every time you push if you're using HTTPS


stbenjam:
Ah neat, and GitHub is smart enough to not ask for password when pulling and username is set.  Looks good to me, ACK

PR #95, opened 2016-12-28
Allow pulp via HTTP

dLobatog:
Anaconda cannot get https content, so the route files should be also allowed on regular HTTP, as in a production installation.


dLobatog:
@ehelms I think the latest update is equivalent to what I had, however I wasn't able to test with lazy-sync as my proxies already have the content & the repo isn't on lazy sync as before.

ehelms:
Thanks @dLobatog !

PR #101, opened 2017-03-14
deploy ktest

sean797:
I've made a few changes to the original `ktest` found https://theforeman.org/plugins/katello/developers.html

This allow you to run it from any directory & using relative paths to test files. e.g
```
[vagrant@centos7-devel test]$ pwd
/home/vagrant/katello/test
[vagrant@centos7-devel test]$ ktest models/content_view_test.rb controllers/api/v2/package_groups_controller_test.rb
``` 

OR

```
[vagrant@centos7-devel models]$ pwd
/home/vagrant/katello/test/models
[vagrant@centos7-devel models]$ ktest content_view_test.rb  -ntest_create
```

### DIFF
--- a/templates/ktest.sh.erb
+++ b/templates/ktest.sh.erb
@@ -1,0 +1,27 @@
+#!/bin/bash
+#
+# Deplyed with katello_devel scenario
+
+FOREMAN_PATH=<%= scope['::katello_devel::deployment_dir'] %>/foreman
+KATELLO_PATH=<%= scope['::katello_devel::deployment_dir'] %>/katello
+
+
+if [[ -n $1 ]]
+then
+  for var in "$@"
+  do
+    if [[ -f $var ]]
+    then
+      TEST_FILES="$TEST_FILES $(readlink -f $var)"
+    else
+      OTHER_OPTS="$OTHER_OPTS $var"
+    fi
+  done
+
+  cd $FOREMAN_PATH
+  RAKE_PATH=`bundle show rake`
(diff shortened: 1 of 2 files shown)


templates/ktest.sh.erb
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env sh
+#
+# Deplyed with katello_devel scenario
+
+FOREMAN_PATH=<%= scope['::katello_devel::deployment_dir'] %>/foreman
+KATELLO_PATH=<%= scope['::katello_devel::deployment_dir'] %>/katello
+
+
+if [[ -n $1 ]]
```
ekohl: This is a bashism. That's fine, but then the shebang should be ```#!/bin/bash```

sean797: Thanks @ekohl I've updated it 👍

templates/ktest.sh.erb
```diff
@@ -0,0 +1,29 @@
...
+
+  cd $FOREMAN_PATH
+  RAKE_PATH=`bundle show rake`
+  ruby -I"lib:test:${KATELLO_PATH}/test:${KATELLO_PATH}/spec" -I"${RAKE_PATH}/lib" $TEST_FILES $OTHER_OPTS
+else
+  cd $FOREMAN_PATH
+  bundle exec rake test:katello
+fi
+
+cd - > /dev/null
```
ekohl: I wonder if this is actually needed. A script has its own working directory, right?

sean797: Yes - you are right, its not required. Updated!

PR #104, opened 2017-03-31
Pass qpid certs to Candlepin

ehelms:


ekohl:
Also needs the qpid exchange variable to ```katello::qpid```.

manifests/init.pp
```diff
@@ -208,10 +210,11 @@
...
   class { '::katello::qpid':
-    client_cert           => $certs::qpid::client_cert,
-    client_key            => $certs::qpid::client_key,
-    katello_user          => $user,
-    candlepin_event_queue => $candlepin_event_queue,
+    client_cert             => $certs::qpid::client_cert,
+    client_key              => $certs::qpid::client_key,
+    katello_user            => $user,
+    candlepin_event_queue   => $candlepin_event_queue,
+    candlepin_qpid_exchange => $::katello::candlepin_qpid_exchange,
```
stbenjam: This is broken on Puppet 4 because nothing is actually loading the katello name space.  Best bet is to provide a value in katello_devel/params.pp I think

ehelms:
Updated

manifests/init.pp
```diff
@@ -90,6 +90,7 @@
 
   $enable_ostree = $katello::params::enable_ostree,
   $candlepin_event_queue = $katello_devel::params::candlepin_event_queue,
+  $candlepin_qpid_exchange = $katello_devel::params::candlepin_qpid_exchange,
```
stbenjam: Need this in params.pp too

ekohl: Since that class inherits from ```katello::params```, does it really?

ehelms: Strange, the candlepin_event_queue isn't in params since it inherits from katello

ehelms: Oh, it just needs docs

stbenjam: Ah right, I forgot we inherit in the params.

manifests/init.pp
```diff
@@ -208,10 +214,11 @@
...
   class { '::katello::qpid':
-    client_cert           => $certs::qpid::client_cert,
-    client_key            => $certs::qpid::client_key,
-    katello_user          => $user,
-    candlepin_event_queue => $candlepin_event_queue,
+    client_cert             => $certs::qpid::client_cert,
+    client_key              => $certs::qpid::client_key,
+    katello_user            => $user,
+    candlepin_event_queue   => $candlepin_event_queue,
+    candlepin_qpid_exchange => $::katello_devel::candlepin_qpid_exchange,
```
ekohl: This variable is also in local scope.

PR #106, opened 2017-05-31
Create commit-msg hook

sean797:
Small thing to notify people earlier about something prprocessor enforces.

Results in something like:

```bash
[vagrant@centos7-devel katello] (sssss)$ git add -A :/ ; git commit -m 'ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss'   -m 'ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss adsssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss'
Commit message is not wrapped at 72nd column
[vagrant@centos7-devel katello] (sssss)$ git commit -m 'sssss'
[sssss a73fe71] sssss
 1 file changed, 1 insertion(+)
[vagrant@centos7-devel katello] (sssss)$ 
```

### DIFF
--- a/manifests/git_repo.pp
+++ b/manifests/git_repo.pp
@@ -28,4 +28,12 @@ define katello_devel::git_repo(
     user     => $katello_devel::user,
   }
 
+  file { "${katello_devel::deployment_dir}/${title}/.git/hooks/commit-msg":
+    ensure  => file,
+    owner   => $::katello_devel::user,
+    group   => $::katello_devel::group,
+    mode    => '0755',
+    content => template('katello_devel/commit-msg.rb.erb'),
+    require => Vcsrepo["${katello_devel::deployment_dir}/${title}"],
+  }
 }
--- a/templates/commit-msg.rb.erb
+++ b/templates/commit-msg.rb.erb
@@ -1,0 +1,16 @@
+#!/usr/bin/env ruby
+
+require 'uri'
+
+file = File.new(ARGV[0], "r")
+while (line = file.gets)
+  if line.chomp.sub(URI.regexp, '').size > 72 && line !~ /^\s{4,}/
+    message = "Commit message is not wrapped at 72nd column\n"
+  end
+end
+file.close
+
+unless message.nil?
+  puts message
+  exit(1)
+end


manifests/git_repo.pp
```diff
@@ -28,4 +28,12 @@
     user     => $katello_devel::user,
   }
 
+  file { "${katello_devel::deployment_dir}/${title}/.git/hooks/commit-msg":
+    ensure  => file,
+    owner   => $::katello_devel::user,
+    group   => $::katello_devel::group,
+    mode    => '0755',
+    content => template('katello_devel/commit-msg.rb.erb'),
+    require => File["${katello_devel::deployment_dir}/${title}"],
```
ekohl: Puppet autorequires parent paths so I think this is redundant.

sean797: Ah that should have been `Vcsrepo` not `File`. I don't think autorequires works with that resource type? 

I'll push it when https://github.com/Katello/puppet-katello_devel/pull/107 is merged. Tests will fail without it.

sean797:
I think https://github.com/brigade/overcommit would be worth it if were going to be adding lots of hooks, but not for 1 IMO

manifests/git_repo.pp
ekohl: You're right, that won't be autorequired.

ekohl:
> I think https://github.com/brigade/overcommit would be worth it if were going to be adding lots of hooks, but not for 1 IMO

That's true, or if it would change regularly.

ehelms:
Will this error out until it's less than 72? We don't always strictly enforce that in all cases.

sean797:
@ehelms Yes - to any git repo that is initialized with this module.

I think we should have 1 standard across the project if we can.

ekohl:
Would an env var to override this be a good compromise?

ehelms:
Yes, that would be fine with me.

@sean797 I don't disagree with standards. However, one problem we face with this slightly older 72 character locking is that our names for entities (e.g. repository, content view, activation key, subscription) are not short and trying to make a legible commit message that describes issues can be hard to fit into 72 without it turning into nonsense.

PR #108, opened 2017-06-02
Refactor apache loading

ekohl:
By using more parameters on the vhost we automatically load the needed modules. This was needed because puppet-katello split and moved fragments to plain files. file() only loads the first existing parameter where template() loads them all.


manifests/apache.pp
```diff
@@ -2,30 +2,72 @@
...
+  concat::fragment { 'katello-ssl-pulp':
+    target  => '05-katello-ssl.conf',
+    order   => 271,
+    content => file('katello/pulp-apache-ssl.conf'),
+  }
+
+  $rewrite_to_https = [
+    {
+      rewrite_cond => [
+        '%{REQUEST_URI} !^\/pulp\/.*',
```
ekohl: I haven't tested it and I wanted to keep this change small, but I think it can also be ```'%{REQUEST_URI} !^\/(pulp|pub|unattended|streamer)(\/.*)?',```

ehelms:
Looks good minus the breaking tests

ekohl:
> Looks good minus the breaking tests

Looks like that's due to a new version of metadata-json-lint

ekohl:
Tests are now green.

PR #115, opened 2017-07-11
Bump puppetlabs-postgresql version and fix syntax in database.pp

jturel:
I was bringing up the centos7-devel image and I was getting this error during the foreman-installer step
`Evaluation Error: Error while evaluating a Resource Statement, Postgresql::Server::Pg_hba_rule[local all]: parameter 'order' expects an Integer value, got String  at /usr/share/katello-installer-base/modules/katello_devel/manifests/database.pp:3
8 on node centos7-devel.example.com`

I was able to fix the problem by changing the 'order' attributes from strings to integers. I also bumped the version of puppetlabs-postgres to be < 6.0.0 from < 5.0.0 which matches puppet-foreman.

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -39,7 +39,7 @@ class katello_devel::database {
       type        => 'local',
       database    => 'all',
       user        => 'all',
-      order       => '001',
+      order       => 1,
       auth_method => 'trust',
     }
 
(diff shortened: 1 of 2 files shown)


dLobatog:
@ekohl This is worthy of a new release and package IMO, it's blocking forklift installs right now. Forklift installs whatever latest version of this module is available.

ekohl:
As discussed on IRC: nightly does take git versions but the pipeline is currently blocked.

PR #120, opened 2017-08-29
Refs #19514 - Add wcache page to katello_devel

chris1984:


manifests/init.pp
```diff
@@ -68,6 +70,7 @@
   Boolean $use_ssh_fork = $katello_devel::params::use_ssh_fork,
   Optional[String] $fork_remote_name = $katello_devel::params::fork_remote_name,
   String $upstream_remote_name = $katello_devel::params::upstream_remote_name,
+  Integer[0, 1000] $qpid_wcache_page_size = $::katello::params::qpid_wcache_page_size,
```
akofink: Could this be `Integer[0, 1000] $qpid_wcache_page_size = $::katello_devel::params::qpid_wcache_page_size` since the katello_devel params are inherited from katello?

manifests/params.pp
```diff
@@ -16,6 +16,8 @@
   $use_rvm = true
   $rvm_ruby = '2.2.4'
 
+  $qpid_wcache_page_size = 4
```
akofink: Is this needed if it's specified in `katello/manifests/params.pp`?

manifests/init.pp
```diff
@@ -170,6 +173,7 @@
     ssl_cert_name          => 'broker',
     ssl_cert_db            => $certs::nss_db_dir,
     ssl_cert_password_file => $certs::qpid::nss_db_password_file,
+    wcache_page_size       => $::katello_devel::params::qpid_wcache_page_size
```
ehelms: Don't reference the params here, otherwise you will never use any value passed by the user.

manifests/init.pp
```diff
@@ -68,6 +70,7 @@
   Boolean $use_ssh_fork = $katello_devel::params::use_ssh_fork,
   Optional[String] $fork_remote_name = $katello_devel::params::fork_remote_name,
   String $upstream_remote_name = $katello_devel::params::upstream_remote_name,
+  Integer[0, 1000] $qpid_wcache_page_size = $::katello_devel::params::qpid_wcache_page_size
```
jturel: Need a comma at the end of this line it seems

ehelms:
You still want to pass that parameter to the qpid:: class.

chris1984:
@ehelms tests are failing and :man_shrugging: not sure why

PR #121, opened 2017-08-31
Refactor

ekohl:
This is the initial commit that refactors the module a bit. It starts to depend more on katello-katello to do the actual backend services installs.

It removes two unused parameters and no longer uses katello::params but instead uses local variables.

The goal is to use katello-katello for every service except the application. For that we continue doing a git install.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -15,15 +15,13 @@
 # $post_sync_token::          The shared secret for pulp notifying katello about
 #                             completed syncs
 #
-# $use_passenger::            Whether to use Passenger in development
-#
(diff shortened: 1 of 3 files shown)


akofink:
:raised_hands: Tested and works great! No need for https://github.com/Katello/puppet-katello_devel/pull/120

ekohl:
Interesting that it works because I had some undefined variables in the settings file. That should work show with the additional commit.

akofink:
The only failure I see is at the `bundle install` step, which you didn't change. I will retry with a fresh environment.

PR #122, opened 2017-09-01
Add option to configure rvm branch

akofink:
RVM's stable branch is broken at the time of opening this PR, but the master/head branch works fine.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -21,6 +21,8 @@
 #
 # $rvm_ruby::                 The default Ruby version to use with RVM
 #
+# $rvm_branch::               The branch to install RVM from; 'stable' or 'head'
+#
 # $qpid_wcache_page_size::    The size (in KB) of the pages in the write page cache
 #
 # $initial_organization::     Initial organization to be created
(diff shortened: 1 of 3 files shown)


ekohl:
Don't you want a parameter in ```init.pp``` so you can actually switch it?

akofink:
Updated. Pardon my lack of puppet knowledge. :apple:

manifests/init.pp
```diff
@@ -54,6 +56,7 @@
   Enum['postgres', 'sqlite'] $db_type = $katello_devel::params::db_type,
   Boolean $use_rvm = $katello_devel::params::use_rvm,
   String $rvm_ruby = $katello_devel::params::rvm_ruby,
+  String $rvm_branch = $katello_devel::params::rvm_branch,
```
ekohl: This works. If it's only stable or head you can enforce it with ```Enum['stable', 'head']``` as well.

akofink:
I don't think we should change the type.

ekohl:
Thanks!

Issue #123, opened 2017-09-14
deployment_url can no longer be passed to candlepin


mdellweg:
https://github.com/Katello/puppet-katello_devel/blob/5ee68ce508526f62797df950e5b2db47ca408d8f/manifests/init.pp#L114

This line breaks forklift devel. I think the problem started with this:
https://github.com/Katello/puppet-candlepin/pull/82

ekohl:
Apologies, I always forget this module. See the related PR for the fix.

PR #124, opened 2017-09-14
Stop passing in deployment_url to katello::candlepin

ekohl:
This parameter was unused and removed.

Fixes GH-123


ekohl:
Now with properly aligned arrows to make puppet-lint happy.

PR #132, opened 2017-10-25
Enable katello in Pulp to install pulp-katello

ehelms:


ekohl:
Can remove it from ```install.pp``` then?

ehelms:
Done. Good catch

PR #134, opened 2017-11-03
include name so plugins that have more than one gemspec don't fail to install

dschlenk:
Trying to add the remote execution plugin to my katello development environment fails because that gem has two gemspec files. Adding the name here seems to fix it without any adverse effects, although I've only got a couple plugins specified in my build. I'd be happy to change it to include a conditional so it only specifies the name attribute for specific plugins but IMHO that smells bad.


ekohl:
I think this should be fine. @akofink mind having a look as well?

akofink:
I've tested it with the forklift devel environment, and it doesn't appear to break anything.

ekohl:
Thanks @dschlenk!

PR #96, opened 2017-01-03
Update development setup

ekohl:
This should make a nightly development setup work again. Inspired by https://github.com/Katello/forklift/pull/351

PR #136, opened 2017-11-06
leave webpack dev server off by default

dschlenk:


akofink:
`webpack_dev_server_https` is false by default already, right?

dschlenk:
@akofink Not sure, but the app works via HTTP on port 3000 and HTTPS via reverse proxy when configured like this. Perhaps `webpack_dev_server_https` doesn't get used if `webpack_dev_server` is false?

xprazak2:
If `webpack_dev_server` is false, then you use the compiled bundle and you do not need the other setting.

akofink:
I don't have merge rights here. @ehelms ?

chris1984:
Thanks @dschlenk

dschlenk:
right on, thanks for the merge @chris1984!

ehelms:
This won't out of the box correct? Since the assets are not compiled? That should perhaps be added if this is the default?

One word of caution, I get the use case here, but as with this issue in the past, I dislike having diverged developer setups due to potential differences in environments.

dschlenk:
@ehelms absolutely could be PEBKAC, but it didn't work when set to true either without following the [webpack ](https://github.com/theforeman/forklift/blob/master/docs/development.md#webpack) instructions so I figured defaulting to false was at least a step shorter to a working box.

ekohl:
https://github.com/theforeman/puppet-katello_devel/pull/96 should properly integrate the webpack setup, but I haven't had time to properly work on it. I like the idea of using the SCL nodejs for a modern nodejs + npm out of the box.

PR #145, opened 2018-02-22
Bundler group names can't handle dashes with symbols

ehelms:
This allows for plugins like foreman-tasks that have a dash since
it cannot handle:

  :foreman-tasks_dev

But can handle:

  'foreman-tasks_dev'


jhutar:
In latest nightly package, katello-installer-base-3.7.0-1.nightly.201802051800gitef10bf4.el7.noarch, this is still not present. Do we know when it will be available? This is obstacle for some QE automation . I understood there is some issue with building nightly packages?

ekohl:
One problem is that https://ci.theforeman.org/job/release_build_katello_installer/ has been failing to build since February 14th with timeouts. That means katello-installer-base-3.7.0-1.nightly.201803071800git874e131.el7.noarch actually contains almost a month old build. The second problem is that our nightly pipeline has been broken for a while which means builds from koji are not being promoted (since the fail tests).

PR #146, opened 2018-02-28
use local vars in git_repo & optional $github_username

ekohl:


ehelms:
We'll need to update the Forklift docs with this change to indicate that fork setup will no longer be default and enforced I think to help existing developers that are used to this behavior.

ekohl:
@ehelms the first commit should be 100% compatible. The second one makes the parameter optional but you can still pass it in. Our docs should still work and until we fix the katello nightly pipeline we don't have the available anyway.

PR #149, opened 2018-03-10
Fix github username still being required

ehelms:
@ekohl I get errors locally if I don;t supply a github_username given both params are then undef.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -79,7 +79,7 @@ class katello_devel (
   Array[String] $extra_plugins = $katello_devel::params::extra_plugins,
 ) inherits katello_devel::params {
 
-  $fork_remote_name_real = pick($fork_remote_name, $github_username)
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -79,7 +79,9 @@
   Array[String] $extra_plugins = $katello_devel::params::extra_plugins,
 ) inherits katello_devel::params {
 
-  $fork_remote_name_real = pick($fork_remote_name, $github_username)
```
ekohl: Maybe just add ```'user'``` as a third option so the variable is always defined?

ehelms:
@ekohl Updated with your suggestion

ekohl:
Turns out [pick_default](https://github.com/puppetlabs/puppetlabs-stdlib#pick_default) is an even better choice. Implemented in https://github.com/theforeman/puppet-katello_devel/pull/150 with tests.

ekohl:
Closing in favor of #150

PR #151, opened 2018-03-15
Ensure systemd-devel is installed

ekohl:


jturel:
Works for me thanks @ekohl - ACK

PR #152, opened 2018-03-23
add SEED_* env vars for the future

sean797:
With this vars set I can more easily recreate my database and run `db:seed` again.


manifests/init.pp
```diff
@@ -183,4 +183,12 @@
     group   => $group,
     mode    => '0755',
   }
+
+  file { '/etc/profile.d/rake-env.sh':
```
ekohl: I'd name this `foreman-rake-env` to be a bit more explicit what it's for.

ehelms:
This is to avoid re-running the installer?

ekohl:
I think this is for when you `db:reset` and `db:migrate db:seed`. It might make more sense to do this in forklift's `foreman_devel` role.

ehelms:
I suppose this is attempting to be overriden, which would make sense:

https://github.com/Katello/katello/blob/master/lib/katello/tasks/reset.rake#L44

PR #153, opened 2018-03-27
Don't setup fork if github_username is blank

ehelms:


manifests/git_repo.pp
```diff
@@ -9,7 +9,7 @@
   Boolean $use_ssh_fork = $katello_devel::use_ssh_fork,
 ) {
 
-  if $github_username {
+  if $github_username != undef or $github_username != '' {
```
ekohl: Shouldn't that be `and` because this is always true?

ehelms: Good call -- updated

PR #158, opened 2018-05-18
Fixes #23648 - point to correct template

beav:
In https://github.com/theforeman/puppet-katello/pull/241, a file was
converted to a template.

This commit changes the same reference in katello_devel.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -38,10 +38,13 @@ class katello_devel::apache {
     request_headers     => ["set X_FORWARDED_PROTO 'https'"],
   }
 
+  # used in template below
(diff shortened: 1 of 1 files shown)


beav:
@ekohl good catch, updated :tropical_fish:

manifests/apache.pp
```diff
@@ -38,10 +38,13 @@
     request_headers     => ["set X_FORWARDED_PROTO 'https'"],
   }
 
+  # used in template below
+  $pub_dir_options = "+FollowSymLinks +Indexes"
```
ekohl: Could you change this to single quotes?

ekohl:
Thanks!

PR #165, opened 2018-07-12
Refs #24220 - increase recommended pool size for the db

iNecas:
Since Rails 5, it seems the webrick is not limited on single request at
a time, which leads to 500s when loading the dasboards, especially with
plugins.

Suggesting to increse the default pool size in the database.yml
to avoid users running into this.


ekohl:
Thanks!

PR #167, opened 2018-07-13
Use puma multi-threaded by default in devel

stbenjam:
So @Izhmash encountered a problem with our dev box with provisioning: when Foreman does an SSH finish template on the host it totally locks up the thread and Foreman won't do anything else (including serving the actual provisioning templates to the host).  This changes our devel config to use puma with 5 workers by default.

I noticed a very big speed-up on a lot of pages by using this configuration, as well.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -63,6 +63,7 @@
 #
 # $extra_plugins::            Array of Github namespace/repo plugins to setup and configure from git
 #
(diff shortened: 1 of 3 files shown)


ekohl:
@stbenjam could you have a look?

stbenjam:
@ekohl Updated

PR #170, opened 2018-08-08
refs #24370 - correct crane config

thomasmckay:
Was incorrectly referencing crane variables in _container_image_registry_ section


templates/katello.yaml.erb
```diff
@@ -30,5 +30,5 @@
 
   # Internal configuration for communication from server to pulp crane service.
   :container_image_registry:
-    :crane_url: <%= @crane_url %>
-    :crane_ca_cert_file: <%= @crane_ca_cert %>
+    :crane_url: <%= scope['katello::crane_url'] %>
```
ekohl: These should be `katello_devel::`

PR #174, opened 2018-09-05
Bump DB pool size

beav:
Awhile back, the DB pool size for development was set to 5. This was not enough, and resulted in `ActiveRecord::ConnectionTimeoutError (could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use)` errors.

The pool size was raised to 10 in https://github.com/theforeman/puppet-katello_devel/pull/165 which helped, but the error would still pop up occasionally when viewing the Katello dashboard.

This commit raises the limit to 20. I have not seen the pool error since raising to 20 on my local environment.

### DIFF
--- a/templates/database.yaml.erb
+++ b/templates/database.yaml.erb
@@ -1,7 +1,7 @@
 development:
   adapter: <%= @db_adapter %>
   database: <%= @db_name %>
-  pool: 10
+  pool: 20
 <% if @db_type == 'sqlite' -%>
   timeout: 5000
 <% else -%>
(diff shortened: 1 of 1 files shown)


akofink:
Is there a reason we don't just increase this to (idk) 100? Do idle pools use significant resources?

Kind of answered my own questions with Google :) It seems like idle connections aren't resource intensive at all, but there is some 'optimal' number, I assume more important for production systems.

https://wiki.postgresql.org/wiki/Number_Of_Database_Connections

beav:
@akofink the main reason to cap it is just so you get an application-level failure on too many connections (out of conns, retry) instead of an os-level failure that may be harder to recover from (out of file handles, different running apps start to crash in unique ways).

My typical rule of thumb is "double it and see what happens" unless there's a reason to select another number, but I have no issue with 50 or 100. IMO let's do 20 and go to 50 if it still happens.

ekohl:
Sounds good. Thanks!

PR #178, opened 2018-10-11
Allow SCL ruby in devel

akofink:
@ekohl Here's my attempt :)

Related to https://github.com/theforeman/forklift/pull/841

### DIFF
--- a/manifests/bundle.pp
+++ b/manifests/bundle.pp
@@ -1,9 +1,11 @@
-# Run a bundle command, possibly under RVM
+# Run a bundle command, possibly under RVM or SCL
 define katello_devel::bundle(
   Array[String] $environment = [],
   $unless = undef,
   $use_rvm = $katello_devel::use_rvm,
   $rvm_ruby = $katello_devel::rvm_ruby,
+  $scl_ruby = $katello_devel::scl_ruby,
+  $use_scl_ruby = $katello_devel::use_scl_ruby,
   $user = $katello_devel::user,
   $cwd = $katello_devel::foreman_dir,
 ) {
@@ -13,6 +15,9 @@ define katello_devel::bundle(
     Class['katello_devel::rvm'] -> Exec["bundle-${title}"]
     $command = "rvm ${rvm_ruby} do bundle ${title}"
     $path = "/home/${user}/.rvm/bin:/usr/bin:/bin"
+  } elsif $use_scl_ruby {
+    $command = "scl enable ${scl_ruby} 'bundle ${title}'"
+    $path = '/usr/bin:/bin'
   } else {
     $command = "bundle ${title}"
     $path = '/usr/bin:/bin'
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -26,6 +26,10 @@
 #
 # $rvm_branch::               The branch to install RVM from; 'stable' or 'head'
 #
+# $use_scl_ruby::             If set to true, will configure with SCL
+#
+# $scl_ruby::                 The default Ruby version to use with SCL
+#
 # $qpid_wcache_page_size::    The size (in KB) of the pages in the write page cache
 #
 # $manage_bundler::           If set to true, will execute the bundler
@@ -75,6 +79,8 @@ class katello_devel (
   Boolean $use_rvm = $katello_devel::params::use_rvm,
   String $rvm_ruby = $katello_devel::params::rvm_ruby,
   String $rvm_branch = $katello_devel::params::rvm_branch,
+  Boolean $use_scl_ruby = $katello_devel::params::use_scl_ruby,
+  String $scl_ruby = $katello_devel::params::scl_ruby,
   Boolean $manage_bundler = $katello_devel::params::manage_bundler,
   String $initial_organization = $katello_devel::params::initial_organization,
   String $initial_location = $katello_devel::params::initial_location,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -1,8 +1,14 @@
 # Katello Development Install
 class katello_devel::install {
 
-  package{ ['libvirt-devel', 'sqlite-devel', 'postgresql-devel', 'libxslt-devel', 'systemd-devel', 'libxml2-devel', 'git', 'npm', 'libcurl-devel']:
+  package{ ['libvirt-devel', 'sqlite-devel', 'postgresql-devel', 'libxslt-devel', 'systemd-devel', 'libxml2-devel', 'git', 'npm', 'libcurl-devel', 'gcc-c++', 'libstdc++']:
     ensure => present,
+  }
+
+  if $katello_devel::use_scl_ruby {
+    package { ["${katello_devel::scl_ruby}-ruby-devel", "${katello_devel::scl_ruby}-rubygem-bundler"]:
+      ensure => present,
+    }
   }
 
   katello_devel::git_repo { 'foreman':
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -15,9 +15,12 @@ class katello_devel::params {
 
   $webpack_dev_server = true
 
-  $use_rvm = true
+  $use_rvm = false
   $rvm_ruby = '2.5'
   $rvm_branch = 'stable'
+
+  $use_scl_ruby = true
+  $scl_ruby = 'rh-ruby25'
 
   $manage_bundler = true
 
--- a/manifests/setup.pp
+++ b/manifests/setup.pp
@@ -22,7 +22,7 @@ class katello_devel::setup (
     }
   }
 
-  katello_devel::bundle { 'install --without mysql:mysql2 --retry 3 --jobs 3':
+  katello_devel::bundle { 'install --without mysql:mysql2 --retry 3 --jobs 3 --path .vendor':
     environment => ['MAKEOPTS=-j'],
   } ->
   katello_devel::bundle { 'exec npm install': } ->


manifests/params.pp
```diff
@@ -19,6 +19,9 @@
   $rvm_ruby = '2.5'
   $rvm_branch = 'stable'
 
+  $use_scl = false
```
ekohl: This needs to be `$use_scl_ruby`

akofink:
I've verified that it doesn't work XD

ehelms:
@akofink What issues do you see with it? I saw you mention Puppet which surprises me (given Puppet is an AIO package). Is there anything else? What do you see fail about Puppet?

akofink:
It seems I was missing some native dependencies (not sure why SCL needs gcc and RVM doesn't, for instance). I've updated this PR with what I've found, but I'm still running into some issues during bundle. Right now, it seems bundle fails when run with the installer but passes when the same command is run manually. :man_shrugging: I'll keep commenting here with further findings.

ehelms:
I hit intaller issues that you mentioned, but when I run the command by hand I legit get a real error:

```
$ scl enable rh-ruby25 'bundle install --without mysql:mysql2 --retry 3 --jobs 3'

There was a ArgumentError while loading katello.gemspec: 
Malformed version number string 3.11.0-master from
  /home/vagrant/katello/katello.gemspec:9:in `block in <main>'
```

manifests/install.pp
```diff
@@ -1,7 +1,7 @@
 # Katello Development Install
 class katello_devel::install {
 
-  package{ ['libvirt-devel', 'sqlite-devel', 'postgresql-devel', 'libxslt-devel', 'systemd-devel', 'libxml2-devel', 'git', 'npm', 'libcurl-devel']:
+  package{ ['libvirt-devel', 'sqlite-devel', 'postgresql-devel', 'libxslt-devel', 'systemd-devel', 'libxml2-devel', 'git', 'npm', 'libcurl-devel', "${katello_devel::scl_ruby}-ruby-devel", 'gcc-c++', 'libstdc++']:
```
ekohl: Could you modify this to be in an if statement and add `rubygem-bundler`?

```puppet
  if $katello_devel::use_scl_ruby {
    package { ["${katello_devel::scl_ruby}-ruby-devel", "${katello_devel::scl_ruby}-rubygem-bundler"]:
      ensure => present,
    }
  }
```

ekohl:
Oh, I also changed the katello version to `3.12.0` while testing. Maybe `3.12.0.master` also works.

akofink:
Updated

ekohl:
Something odd now happens: bundle install finishes successfully but then on npm install it fails because nokogiri is not in the bundle. A later (manual) bundle install shows the same thing.

ekohl:
This is the GEM_HOME issue in the rh-ruby25 SCL. https://bugzilla.redhat.com/show_bug.cgi?id=1613864#c5 is a workaround. `bundle install --path .vendor` might also work around this.

johnpmitsch:
I hit the malformed string error, once I got past that by manually changing the string (haven't figured out a real fix), I get errors about missing gems on `npm install`. I suspect that the issue is the npm post install scripts use a child process to run things [here](https://github.com/theforeman/foreman/blob/c18d67464d81943b77feeb0e24ec649574860ab7/script/npm_install_plugins.js#L5-L9) and it doesn't have the correct environment. I'll continue to look into things

manifests/bundle.pp
```diff
@@ -13,6 +15,9 @@
     Class['katello_devel::rvm'] -> Exec["bundle-${title}"]
     $command = "rvm ${rvm_ruby} do bundle ${title}"
     $path = "/home/${user}/.rvm/bin:/usr/bin:/bin"
+  } elsif $use_scl_ruby {
+    $command = "scl enable ${scl_ruby} 'bundle ${title} --path .vendor'"
```
ehelms: Drop the `--path .vendor` here as this breaks `bundle exec`. Instead move it to here -- https://github.com/theforeman/puppet-katello_devel/blob/master/manifests/setup.pp#L25

I got a green install making that change.

akofink:
Updated - thanks for the help testing this PR!

ekohl:
https://github.com/theforeman/puppet-katello_devel/pull/193 should fix the tests btw.

iNecas:
#193 is merged, let's rebase to get the green tests

manifests/init.pp
```diff
@@ -26,6 +26,10 @@
 #
 # $rvm_branch::               The branch to install RVM from; 'stable' or 'head'
 #
+# $use_scl_ruby::             If set to true, will configure with SCL
+#
+# $scl_ruby::                 The default Ruby version to use with RVM
```
iNecas: `s/RVM/SCL/`

akofink:
Rebased!

akofink:
Anyone else want to block this? Or is it good to merge?

manifests/params.pp
```diff
@@ -19,6 +19,9 @@
   $rvm_ruby = '2.5'
   $rvm_branch = 'stable'
 
+  $use_scl_ruby = false
```
johnpmitsch: just one question, should we default to true? Seems like the plan is to move the dev box to the ruby SCL rather than RVM

ekohl: I'm thinking about this. How would the migration path be?

akofink: I'd say there's no migration path. It's a devel box, so I don't feel too bad about that. If we default to true, I think we just need to post to the community forum to show people how to keep RVM turned on for existing devel boxes.

manifests/params.pp
```diff
@@ -15,10 +15,13 @@
 
   $webpack_dev_server = true
 
-  $use_rvm = true
+  $use_rvm = false
   $rvm_ruby = '2.5'
   $rvm_branch = 'stable'
 
+  $use_scl_ruby = true
+  $scl_ruby = 'rh-ruby25'
```
akofink: I went ahead and defaulted to SCL.

manifests/params.pp
ekohl: A reinstall is also a migration path ;) But a post to the forum is good. We will get questions about this and it's easy to refer to on IRC.

johnpmitsch: ya, if you have a dev box with RVM that is fine. RVM is mostly an issue when provisioning boxes, once they are functional things are generally fine. I wouldn't worry too much about existing boxes (unless I'm missing something).

So we should set the SCL to true here? :pray:

johnpmitsch:
Doing one last test!

johnpmitsch:
Ok, I may have spoke too soon, sorry!

I do see this when starting the server, this is after I re-ran the installer and it succeeded

```
[vagrant@centos7-devel-scl foreman]$ bundle exec foreman start
There was a ArgumentError while loading katello.gemspec: 
Malformed version number string 3.12.0-master from
  /home/vagrant/katello/katello.gemspec:9:in `block in <main>'
```

johnpmitsch:
Seems like ruby 2.0.0 is being used too?

```
[vagrant@centos7-devel-scl foreman]$ bundle install
Your Gemfile lists the gem factory_bot_rails (~> 4.5) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
Your Gemfile lists the gem webmock (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
Fetching source index from https://rubygems.org/
Resolving dependencies........
Installing CFPropertyList 2.3.6
Installing rake 12.3.2
Installing minitest 5.10.3
Installing concurrent-ruby 1.1.4
Installing builder 3.2.3
Installing thread_safe 0.3.6
Installing erubi 1.8.0
Installing mini_portile2 2.4.0

Gem::InstallError: rack requires Ruby version >= 2.2.2.
Installing crass 1.0.4
An error occurred while installing rack (2.0.6), and Bundler cannot continue.
Make sure that `gem install rack -v '2.0.6'` succeeds before bundling.
[vagrant@centos7-devel-scl foreman]$ ruby -v
ruby 2.0.0p648 (2015-12-16) [x86_64-linux]
```

akofink:
I ran into the same error on install, and a re-install passed just like yours, but my server comes up fine, and my ruby is 2.5.3 from the SCL:

```
[vagrant@centos7-devel-test foreman]$ which ruby
/opt/rh/rh-ruby25/root/usr/bin/ruby
[vagrant@centos7-devel-test foreman]$ which bundle
/opt/rh/rh-ruby25/root/usr/bin/bundle
[vagrant@centos7-devel-test foreman]$ bundle exec ruby -v
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-linux]
[vagrant@centos7-devel-test foreman]$ ruby -v
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-linux]
[vagrant@centos7-devel-test foreman]$ bundle exec rails s -b 0.0.0.0
/home/vagrant/foreman/lib/foreman.rb:8: warning: already initialized constant Foreman::UUID_REGEXP
/home/vagrant/foreman/lib/foreman.rb:8: warning: previous definition of UUID_REGEXP was here
=> Booting Puma
=> Rails 5.2.1 application starting in development 
=> Run `rails server -h` for more startup options
The Apipie cache is turned off. Enable it and run apipie:cache rake task to speed up API calls.
/home/vagrant/foreman/app/services/facets.rb:2: warning: already initialized constant Facets::SUPPORTED_CORE_OBJECTS
/home/vagrant/foreman/app/services/facets.rb:2: warning: previous definition of SUPPORTED_CORE_OBJECTS was here
2019-03-07T16:43:23 [D|app|] Current user set to foreman_admin (admin)
2019-03-07T16:43:23 [D|app|] Current user set to foreman_admin (admin)
2019-03-07T16:43:23 [D|app|] Registering 4 assets for plugin foreman-tasks precompilation
2019-03-07T16:43:23 [D|app|] Registering 4 assets for plugin foreman_docker precompilation
2019-03-07T16:43:23 [D|app|] Registering 114 assets for plugin bastion precompilation
2019-03-07T16:43:23 [W|app|] Creating scope :completer_scope. Overwriting existing method Organization.completer_scope.
2019-03-07T16:43:24 [W|app|] Scoped order is ignored, it's forced to be batch order.
2019-03-07T16:43:24 [W|app|] Creating scope :completer_scope. Overwriting existing method Location.completer_scope.
2019-03-07T16:43:25 [D|app|] Current user set to foreman_admin (admin)
2019-03-07T16:43:25 [D|app|] Registering 100 assets for plugin katello precompilation
2019-03-07T16:43:25 [W|app|] ignoring associations organization_ids, location_ids audit definition for Container, the resource is not audited
2019-03-07T16:43:26 [W|app|] DEPRECATION WARNING: `secrets.secret_token` is deprecated in favor of `secret_key_base` and will be removed in Rails 6.0. (called from <main> at /home/vagrant/foreman/config/environment.rb:5)
2019-03-07T16:43:26 [W|app|] DEPRECATION WARNING: Using a dynamic :controller segment in a route is deprecated and will be removed in Rails 6.0. (called from block in <main> at /home/vagrant/foreman/config/routes.rb:17)
Puma starting in single mode...
* Version 3.12.0 (ruby 2.5.3-p105), codename: Llamas in Pajamas
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
```

johnpmitsch:
trying once more again! I was behind on some forklift commits so that could be the difference

akofink:
https://community.theforeman.org/t/incoming-change-scl-ruby-by-default-in-forklift-centos7-devel-box/13181

PR #180, opened 2018-10-17
Fix ktest single mode

mdellweg:
Due to 3784e5bc9f17c38a3cde9bf2c5e458a88398e5da we need `bundler exec` to
run ruby now.

Thanks to @leewaa, @ekohl and @timogoebel for investigation and fixing.


ekohl:
Thanks!

PR #193, opened 2019-02-11
Correct the puppet-squid fixture

ekohl:
puppet-pulp switched its squid implementation so we follow suit.


ehelms:
Looks like this needs to reference `selinux` module as well.

ekohl:
:green_heart:

PR #195, opened 2019-03-12
Switch to rh-nodejs8 for node runtime

ehelms:
If you are wondering why not default to nodejs 10, this bug is your answer:

https://bugzilla.redhat.com/show_bug.cgi?id=1677710


ehelms:
Should go alongside https://github.com/theforeman/forklift/pull/934

ehelms:
We also use nodejs 6 currently in our build environment so this change may not be wanted due to that fact.

ekohl:
> We also use nodejs 6 currently in our build environment so this change may not be wanted due to that fact.

I would not be opposed to switch the build env as well. Especially now that nodejs 6 has been removed from the Travis config. This may be easier if we start using `@theforeman/vendor` since then we have less packages to convert.

PR #196, opened 2019-03-14
Remove --path vendor from bundle commands

akofink:
This should improve performance.


akofink:
@johnpmitsch or @ekohl care to review?

PR #204, opened 2019-05-31
Revert "Remove --path vendor from bundle commands"

ekohl:
Reverts theforeman/puppet-katello_devel#196

This doesn't work right now. Probably because the Puppet exec doesn't read `GEM_HOME` when running `bundle install`.


ekohl:
@akofink could you have a look?

PR #205, opened 2019-06-03
Make pkill not fail on error

johnpmitsch:
Just testing this out to unblock this error and see what its hiding :space_invader: 

No need to merge unless you think its a good change to have permanently

### DIFF
--- a/manifests/setup.pp
+++ b/manifests/setup.pp
@@ -38,7 +38,7 @@ class katello_devel::setup (
     } ->
     Class['foreman_proxy::register'] ->
     exec { 'destroy rails server':
-      command   => "/usr/bin/pkill -9 --pidfile ${pidfile}",
+      command   => "/usr/bin/pkill -9 --pidfile ${pidfile} || true",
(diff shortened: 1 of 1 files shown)


ekohl:
How was the experiment? Can we close this PR?

johnpmitsch:
@ekohl Thanks for the poke! This actually cleared up the error for https://github.com/theforeman/forklift/pull/941 - I'm not sure why I'm only seeing this command error with packer. I still have more work to do on the forklift PR, the packer box isn't functioning as expected.

If this is a harmless change, maybe we should merge it? I think the PID not existing shouldn't mean the installer errors out, but then again, maybe it's symptomatic of some other issue.

johnpmitsch:
from an off-thread discussion:

> you can also disable the registration (--foreman-proxy-register-in-foreman false from the top of my head) which avoids the whole start + kill

johnpmitsch:
Closing this out until I have more time to investigate the packer PR

PR #208, opened 2019-09-09
Add timeout to npm install

johnpmitsch:
We are seeing npm install timeout in multiple katello dev environment runs

### DIFF
--- a/manifests/bundle.pp
+++ b/manifests/bundle.pp
@@ -8,6 +8,7 @@ define katello_devel::bundle(
   $use_scl_ruby = $katello_devel::use_scl_ruby,
   $user = $katello_devel::user,
   $cwd = $katello_devel::foreman_dir,
+  Integer[0] $timeout = 600,
 ) {
 
   if $use_rvm {
@@ -29,7 +30,7 @@ define katello_devel::bundle(
     cwd         => $cwd,
     user        => $user,
     logoutput   => 'on_failure',
-    timeout     => '600',
+    timeout     => $timeout,
     path        => $path,
     unless      => $unless,
   }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -68,6 +68,9 @@
 # $extra_plugins::            Array of Github namespace/repo plugins to setup and configure from git
 #
 # $rails_command::            Customize the command used to start rails
+#
+# $npm_timeout::              Timeout for npm install step
+#
 class katello_devel (
   String $user = $katello_devel::params::user,
   Stdlib::Absolutepath $deployment_dir = $katello_devel::params::deployment_dir,
@@ -100,6 +103,7 @@ class katello_devel (
   Integer[0, 1000] $qpid_wcache_page_size = $katello_devel::params::qpid_wcache_page_size,
   Array[String] $extra_plugins = $katello_devel::params::extra_plugins,
   String $rails_command = $katello_devel::params::rails_command,
+  Integer[0] $npm_timeout = $katello_devel::params::npm_timeout,
 ) inherits katello_devel::params {
 
   $fork_remote_name_real = pick_default($fork_remote_name, $github_username)
--- a/manifests/setup.pp
+++ b/manifests/setup.pp
@@ -6,6 +6,7 @@ class katello_devel::setup (
   $initial_location = $katello_devel::initial_location,
   $admin_password = $katello_devel::admin_password,
   $webpack_dev_server = $katello_devel::webpack_dev_server,
+  $npm_timeout = $katello_devel::npm_timeout,
 ) {
   $pidfile = "${foreman_dir}/tmp/pids/server.pid"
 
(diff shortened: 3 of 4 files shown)


manifests/setup.pp
```diff
@@ -25,7 +25,9 @@
   katello_devel::bundle { 'install --without mysql:mysql2 --retry 3 --jobs 3 --path .vendor':
     environment => ['MAKEOPTS=-j'],
   } ->
-  katello_devel::bundle { 'exec npm install': } ->
+  katello_devel::bundle { 'exec npm install':
+    timeout => '600'
```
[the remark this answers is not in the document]
ekohl: Looks like I had a draft review that I forgot to submit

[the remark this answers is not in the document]
johnpmitsch: I'm open to it, but if it is just the devel installer, I wonder how configurable we want it to be? I would think we just want to give npm install enough time to complete and devs shouldn't really care about a timeout but maybe I'm missing something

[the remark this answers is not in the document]
ekohl: My experience is that you almost always run into the point where you want to change it. For example, you need it to be 700. Making it a class parameter allows you to easily tune it without changing the source. So more of a good practice that I've learned to use the hard way.

johnpmitsch:
@ekohl I updated to use a parameter, I'm not too familiar with puppet so hopefully this is what you meant :) I also updated the default to be 900, which is 15 minutes

manifests/init.pp
```diff
@@ -100,12 +100,15 @@
   Integer[0, 1000] $qpid_wcache_page_size = $katello_devel::params::qpid_wcache_page_size,
   Array[String] $extra_plugins = $katello_devel::params::extra_plugins,
   String $rails_command = $katello_devel::params::rails_command,
+  Integer[0, 3600] $npm_timeout = $katello_devel::params::npm_timeout,
```
ekohl: You don't need to set an upper limit

```suggestion
  Integer[0] $npm_timeout = $katello_devel::params::npm_timeout,
```

manifests/init.pp
```diff
@@ -100,12 +100,15 @@
...
 ) inherits katello_devel::params {
 
   $fork_remote_name_real = pick_default($fork_remote_name, $github_username)
 
   $foreman_dir = "${deployment_dir}/foreman"
 
+  $npm_timeout = 900
```
ekohl: This should be in params.pp

johnpmitsch:
@ekohl thanks, updated!

manifests/setup.pp
```diff
@@ -25,7 +26,9 @@
   katello_devel::bundle { 'install --without mysql:mysql2 --retry 3 --jobs 3 --path .vendor':
     environment => ['MAKEOPTS=-j'],
   } ->
-  katello_devel::bundle { 'exec npm install': } ->
+  katello_devel::bundle { 'exec npm install':
+    timeout => $npm_timeout,
```
ekohl: Test failure is relevant:

```
       Evaluation Error: Error while evaluating a Resource Statement, Katello_devel::Bundle[exec npm install]: has no parameter named 'timeout' (file: /home/travis/build/theforeman/puppet-katello_devel/spec/fixtures/modules/katello_devel/manifests/setup.pp, line: 29) on node travis-job-d9b6a10e-5bb7-499e-a206-ee586ce9ecf7.c.travis-ci-prod-2.internal
```

Looks like you'll need to add the parameter to https://github.com/theforeman/puppet-katello_devel/blob/master/manifests/bundle.pp

johnpmitsch:
@ekohl this is getting pretty complicated for one parameter

> It could have been sufficient to only add it to setup.pp 

Is this a better approach?

ekohl:
> @ekohl this is getting pretty complicated for one parameter
> 
> > It could have been sufficient to only add it to setup.pp
> 
> Is this a better approach?

You already have the code now, but it would have been easier to do with in the end sufficient flexibility. You wouldn't have needed to add it to `init.pp` and `params.pp`.

manifests/bundle.pp
```diff
@@ -8,6 +8,7 @@
   $use_scl_ruby = $katello_devel::use_scl_ruby,
   $user = $katello_devel::user,
   $cwd = $katello_devel::foreman_dir,
+  $npm_timeout = $katello_devel::npm_timeout,
```
ekohl:
```suggestion
  Integer[0] $timeout = 600,
```

You'll also need to change line 33 and and use it there

manifests/bundle.pp
```diff
@@ -29,8 +30,9 @@
     cwd         => $cwd,
     user        => $user,
     logoutput   => 'on_failure',
-    timeout     => '600',
+    timeout     => $timeout,
     path        => $path,
     unless      => $unless,
+    npm_timeout => $npm_timeout
```
ekohl: This does not exist: https://puppet.com/docs/puppet/latest/type.html#exec-attributes
```suggestion
```

johnpmitsch:
Looks like the failures are all the `::ipaddress` tests now

ekohl:
https://github.com/theforeman/puppet-foreman_proxy/pull/537 should fix that.

ekohl:
Restart the build now that that's merged.

PR #210, opened 2019-10-23
Switch to using NodeJS from SCL

ehelms:


ehelms:
Yessir -- I was already testing that :) -- https://github.com/theforeman/forklift/pull/1038

PR #216, opened 2019-12-06
initial pulp3 support

wbclark:
Adding the option to install pulp3. Installation is handled by puppet-pulpcore with `manage_apache => false`. A template is used to add `ProxyPass{,Reverse}` declarations to `05-katello.conf`.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -60,6 +60,17 @@ class katello_devel::apache {
     content => template('katello/pulp-apache-ssl.conf.erb'),
   }
 
+  if $katello_devel::pulp3 {
+    class { 'pulpcore':
+      manage_apache => false,
+    }
+
+    concat::fragment { 'katello-pulpcore':
+      target  => '05-katello.conf',
+      content => template('katello_devel/pulpcore-apache.conf.erb'),
+    }
+  }
+
   $rewrite_to_https = [
     {
       rewrite_cond => [
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -51,4 +51,6 @@ class katello_devel::params {
   $rails_command='puma -w 5 -p $PORT --preload'
 
   $npm_timeout = 900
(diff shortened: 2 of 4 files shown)


manifests/apache.pp
```diff
@@ -60,6 +60,17 @@
     content => template('katello/pulp-apache-ssl.conf.erb'),
   }
 
+  if $katello_devel::pulp3 {
+    class { 'pulpcore':
+      manage_apache => false,
+    }
+
+    concat::fragment { 'katello-pulpcore':
+      target  => '05-katello.conf',
```
ekohl: This doesn't exist. You can also use a simpler way: add it to $proxy_pass_https above.

manifests/params.pp
```diff
@@ -51,4 +51,6 @@
   $rails_command='puma -w 5 -p $PORT --preload'
 
   $npm_timeout = 900
+
+  $pulp3 = false
```
ekohl: I'd default to true since every developer should have Pulp 3. @ehelms also suggested to stick to `pulpcore` instead of `pulp3`.

manifests/apache.pp
```diff
@@ -60,6 +60,17 @@
     content => template('katello/pulp-apache-ssl.conf.erb'),
   }
 
+  if $katello_devel::pulp3 {
+    class { 'pulpcore':
```
ekohl: Including `pulpcore` in the Apache config doesn't feel correct. I'd do it in `init.pp`.

ekohl:
By now there is support in puppet-foreman_proxy_content:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/45e251bd10715cbc0d8fce4be67b0897e269b503/manifests/init.pp#L337-L403

Is it time to rebase this PR? I don't know how much of it should be reused vs just calling puppet-pulpcore directly.

PR #224, opened 2020-02-12
Refs #28924 - drop Qpid, add Artemis

jturel:


templates/katello.yaml.erb
```diff
@@ -20,14 +20,16 @@
     :oauth_secret: <%= scope['katello_devel::oauth_secret'] %>
     :ca_cert_file: <%= scope['katello_devel::candlepin_ca_cert'] %>
 
+  :candlepin_events:
+    :broker_host: <%= scope['katello_devel::candlepin_events_broker_host'] %>
+    :broker_port: <%= scope['katello_devel::candlepin_events_broker_port'] %>
+    :queue_name: <%= scope['katello_devel::candlepin_events_queue_name'] %>
+    :subscription_name: <%= scope['katello_devel::candlepin_events_subscription_name'] %>
```
ehelms: This is the pattern we've previously always taken. I've been starting to question this in the present. Are these configurations that need to be done and made prior to runtime? And, do all of them need to be *configurable*?

jturel: > Are these configurations that need to be done and made prior to runtime?

IMO the app should be told all of these values

> And, do all of them need to be _configurable_?

I don't feel strongly about it and was mostly anticipating 'make these configurable' feedback!

ehelms: I'll break it down a bit further. I see it as two types of configuration:

 1) configuration that the application needs to start **and** that should require an application restart to change
 2) configuration that the application needs to execute workflows but should not break the application if its missing or mis-configured

Given your worker with the listener, it's not clear to me where `broker_host` and `broker_port` fit into this model. The latter two options `queue_name` and `subscription_name` feel like values that ought to be hard-coded rather than configurable by a user.

jturel: For me it comes down to this question: whose responsibility is it to know the queue name? The listener is not much more than code that subscribes and handles events it receives. I don't think it should know anything about its queue.

My final PR shows how in engine.rb the CandlepinEventListener is given a connection to subscribe on: https://github.com/Katello/katello/pull/8563/files#diff-cf74e596d7887011413ee9d06b16b753R125-R132

Another thing to consider is that the queue and subscription name are in the broker.xml and they should match Katello's configuration. It makes sense to me for their values to originate from somewhere outside the code - a single place. I think even the client_id at the above link should probably be moved out since it also exists in broker.xml.

The presence of a value in a config file is not always for the user's sake. In this case *we* want to configure the connection details.

jturel: Since my last comment I saw my tests were failing on the katello PR, so I added default settings in engine.rb https://github.com/Katello/katello/pull/8563/files#diff-cf74e596d7887011413ee9d06b16b753R52-R57

This means that the values are still derived from the configuration but we don't need to supply them in dev.

But my earlier point is still relevant - these should be configurable because broker.xml and the config settings need to be in agreement.

ehelms:
@ekohl Do these errors stem from the voxpupuli changes? Whats the remediation?

ekohl:
A PR to make it clean again. For most lint failures, `bundle exec rake lint_fix` should do the job. https://github.com/theforeman/puppet-katello_devel/pull/233

jturel:
@ekohl thank you - now we're green!

cc @ehelms

templates/katello.yaml.erb
```diff
@@ -18,14 +18,15 @@
     :oauth_secret: <%= scope['katello_devel::oauth_secret'] %>
     :ca_cert_file: <%= scope['katello_devel::candlepin_ca_cert'] %>
 
+  :candlepin_events:
+    :ssl_cert_file: <%= scope['certs::candlepin::client_cert'] %>
+    :ssl_key_file: <%= scope['certs::candlepin::client_key'] %>
+    :ssl_ca_file: <%= scope['katello_devel::candlepin_ca_cert'] %>
```
ekohl, line 24: I think we should introduce `$certs::candlepin::ca_cert` for consistency. Will see if I can come up with a PR.

templates/katello.yaml.erb
```diff
@@ -18,14 +18,15 @@
     :oauth_secret: <%= scope['katello_devel::oauth_secret'] %>
     :ca_cert_file: <%= scope['katello_devel::candlepin_ca_cert'] %>
 
+  :candlepin_events:
+    :ssl_cert_file: <%= scope['certs::candlepin::client_cert'] %>
```
ekohl, line 22: Just thinking out loud: this feels wrong. The class `certs::candlepin` should be executed on the candlepin server. A katello server doesn't have to be the same, so this class may not be included. In the current devel module it's not that relevant, but for `puppet-katello` it's incorrect.

templates/katello.yaml.erb
ekohl: Looks like this variable is already available, but it's a different file.

IMHO we should introduce a `certs::candlepin_client` class and katello should use that.

jturel:
@ekohl  I don't disagree with any of your feedback but can these things be addressed later? the puppet-katello changes are already merged and this is a devel module after all.

johnpmitsch:
@jturel Is this change related at all to the current dev box failure?

```
    centos7-katello-devel-stable: TASK [foreman_installer : Run installer] ***************************************
    centos7-katello-devel-stable: fatal: [default]: FAILED! => changed=true
    centos7-katello-devel-stable:   cmd: |-
    centos7-katello-devel-stable:     foreman-installer -v --no-colors --disable-system-checks --scenario katello-devel  --katello-devel-scl-ruby=rh-ruby25 --katello-devel-admin-password changeme
    centos7-katello-devel-stable:   delta: '0:00:04.028053'
    centos7-katello-devel-stable:   end: '2020-04-24 13:35:36.895243'
    centos7-katello-devel-stable:   msg: non-zero return code
    centos7-katello-devel-stable:   rc: 22
    centos7-katello-devel-stable:   start: '2020-04-24 13:35:32.867190'
    centos7-katello-devel-stable:   stderr: |-
    centos7-katello-devel-stable:     tput: unknown terminal "unknown"
    centos7-katello-devel-stable:     [[37m[41mFATAL[0m [34m2020-04-24T13:35:36[0m [36mfatal[0m] Unable to continue because of: documentation mentioned unknown parameters in katello_devel: candlepin_event_queue, candlepin_qpid_exchange
    centos7-katello-devel-stable:   stderr_lines: <omitted>
    centos7-katello-devel-stable:   stdout: ''
    centos7-katello-devel-stable:   stdout_lines: <omitted>
```

ekohl:
Yes, this should have removed the docs. I'll have to look into why our CI didn't catch this. Perhaps fallout from my cleanups.

@jturel or @johnpmitsch mind sending a PR to remove the parameter documentation?

jturel:
Thanks - I opened https://github.com/theforeman/puppet-katello_devel/pull/235

PR #225, opened 2020-02-14
Use foreman apache configs to unblock Pulp 3

ehelms:
This module had declared it's own vhost using a similar naming convention and directory structure as Foreman's. This causes a conflict when pulpcore is enabled since pulpcore relies on adding fragments to Foreman's vhost declaration specifically around the use of `05-foreman-ssl.d`. To solve this, and bring greater continuity I have attempted to declare the `foreman::config::apache` class and re-use it here.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -1,100 +1,18 @@
 # Setups Apache for Katello development
 class katello_devel::apache {
 
-  include apache
-
-  $priority = '05'
-
-  $proxy_pass_https = [
-    {
-      'no_proxy_uris' => ['/pulp', '/streamer', '/pub'],
-      'path'          => '/',
-      'url'           => "http://localhost:${katello_devel::rails_port}/",
-      'params'        => {'retry' => '0'},
-    },
-    {
-      'path'          => '/',
-      'url'           => 'http://localhost:6006/',
-      'params'        => {'retry' => '0'},
-    },
-  ]
-
-  file { "${apache::confd_dir}/${priority}-foreman-ssl.d":
-    ensure  => 'directory',
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0644',
-    purge   => true,
-    recurse => true,
+  class { 'foreman::config::apache':
+    passenger     => false,
+    ssl           => true,
+    ssl_cert      => $certs::apache::apache_cert,
+    ssl_key       => $certs::apache::apache_key,
+    ssl_ca        => $certs::apache::ca_cert,
+    ssl_chain     => $certs::apache::apache_ca_cert,
+    proxy_backend => "http://localhost:${katello_devel::rails_port}/",
   }
 
-  apache::vhost { 'katello-ssl':
-    servername            => $certs::apache::hostname,
-    serveraliases         => $certs::apache::cname,
-    docroot               => '/var/www',
-    port                  => 443,
-    priority              => $priority,
-    options               => ['SymLinksIfOwnerMatch'],
-    ssl                   => true,
-    ssl_cert              => $certs::apache::apache_cert,
-    ssl_key               => $certs::apache::apache_key,
-    ssl_ca                => $certs::apache::ca_cert,
-    ssl_verify_client     => 'optional',
-    ssl_options           => '+StdEnvVars',
-    ssl_verify_depth      => '3',
-    custom_fragment       => file('katello/katello-apache-ssl.conf'),
-    additional_includes   => ["${apache::confd_dir}/${priority}-foreman-ssl.d/*.conf"],
-    use_optional_includes => true,
-    ssl_proxyengine       => true,
-    proxy_pass            => $proxy_pass_https,
-    proxy_preserve_host   => true,
-    request_headers       => ["set X_FORWARDED_PROTO 'https'"],
-  }
-
-  # used in template below
-  $pub_dir_options = '+FollowSymLinks +Indexes'
-
-  concat::fragment { 'katello-ssl-pulp':
-    target  => '05-katello-ssl.conf',
-    order   => 271,
-    content => template('katello/pulp-apache-ssl.conf.erb'),
-  }
-
-  $rewrite_to_https = [
-    {
-      rewrite_cond => [
-        '%{REQUEST_URI} !^\/pulp\/.*',
-        '%{REQUEST_URI} !^\/pulp$',
-        '%{REQUEST_URI} !^\/pub\/.*',
-        '%{REQUEST_URI} !^\/pub$',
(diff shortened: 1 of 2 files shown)


manifests/apache.pp
```diff
@@ -1,101 +1,46 @@
...
+    ssl_protocol            => undef,
+    ssl_verify_client       => 'optional',
+  }
 
-  $priority = '05'
+  foreman::config::apache::fragment { 'katello':
+    ssl_content => file('katello/katello-apache-ssl.conf'),
+  }
 
   $proxy_pass_https = [
```
ehelms: @ekohl I wasn't sure how to declare this additional proxy pass information to the class.

ehelms:
@ekohl general thoughts on this overall approach? It is related to this change https://github.com/theforeman/forklift/pull/1105

manifests/apache.pp
```diff
@@ -1,101 +1,46 @@
...
+		server_ssl_port         => $foreman::params::server_ssl_port,
+		ssl_certs_dir           => '',
+		ssl_crl                 => '',
+		user                    => $foreman::params::user,
+		passenger_prestart      => false,
+		passenger_min_instances => 0,
+		passenger_start_timeout => 0,
+		foreman_url             => $foreman::params::foreman_url,
+		ipa_authentication      => false,
+    proxy_backend           => "http://0.0.0.0:${katello_devel::rails_port}/",
```
ekohl:
```suggestion
    proxy_backend           => "http://localhost:${katello_devel::rails_port}/",
```

manifests/apache.pp
```diff
@@ -1,101 +1,46 @@
...
+  include foreman::params
+
+  class { 'foreman::config::apache':
+    passenger               => false,
+    ssl                     => true,
+    ssl_cert                => $certs::apache::apache_cert,
+    ssl_key                 => $certs::apache::apache_key,
+    ssl_ca                  => $certs::apache::ca_cert,
+    ssl_chain               => $certs::apache::ca_cert,
+		app_root                => '/home/vagrant/foreman',
```
ekohl: Looks like tabs instead of spaces

manifests/apache.pp
```diff
@@ -1,101 +1,46 @@
...
+		serveraliases           => $foreman::params::serveraliases,
+		server_port             => $foreman::params::server_port,
+		server_ssl_port         => $foreman::params::server_ssl_port,
+		ssl_certs_dir           => '',
+		ssl_crl                 => '',
+		user                    => $foreman::params::user,
+		passenger_prestart      => false,
+		passenger_min_instances => 0,
+		passenger_start_timeout => 0,
+		foreman_url             => $foreman::params::foreman_url,
```
ekohl:
```suggestion
    foreman_url             => "https://${certs::apache::hostname}/",
```

manifests/apache.pp
ekohl: I actually wonder if this is really needed since I'm not sure what listens on port 6006. Looks like the storybook according to 3a61896824f67b050f4ce31467adea1e1f7e5b7c but perhaps the devs should just go to `http://$hostname:6006` for that.

ekohl:
I'd like to see a version that uses https://github.com/theforeman/puppet-foreman/pull/800 because it should be simpler.

manifests/apache.pp
ehelms: Developers still access the dev server directly, which means it does not run on localhost.

ekohl: I hope they let it listen on `*` or `0.0.0.0` which means it still listens on localhost. I also wonder why they need to access the dev server directly if Apache is in front of it.

manifests/apache.pp
```diff
@@ -1,100 +1,18 @@
...
-    proxy_pass            => $proxy_pass_https,
-    proxy_preserve_host   => true,
-    request_headers       => ["set X_FORWARDED_PROTO 'https'"],
+  class { 'foreman::config::apache':
+    passenger     => false,
+    ssl           => true,
+    ssl_cert      => $certs::apache::apache_cert,
+    ssl_key       => $certs::apache::apache_key,
+    ssl_ca        => $certs::apache::ca_cert,
+		app_root      => '/home/vagrant/foreman',
```
ekohl: This isn't needed
```suggestion
```

manifests/apache.pp
```diff
@@ -1,100 +1,18 @@
...
+    proxy_backend => "http://0.0.0.0:${katello_devel::rails_port}/",
```
ekohl: I still think this can just be localhost

manifests/apache.pp
```diff
@@ -1,100 +1,18 @@
...
-    ssl_proxyengine       => true,
-    proxy_pass            => $proxy_pass_https,
-    proxy_preserve_host   => true,
-    request_headers       => ["set X_FORWARDED_PROTO 'https'"],
+  class { 'foreman::config::apache':
+    passenger     => false,
+    ssl           => true,
+    ssl_cert      => $certs::apache::apache_cert,
+    ssl_key       => $certs::apache::apache_key,
+    ssl_ca        => $certs::apache::ca_cert,
```
ekohl: If we want to mirror the production setup (which supports custom certs), should this also set the chain? I believe this should fix it.
```suggestion
    ssl_ca        => $certs::apache::ca_cert,
    ssl_chain     => $certs::apache::apache_ca_certs,
```

manifests/init.pp
```diff
@@ -150,7 +150,7 @@
 
   if $manage_bundler {
     class { 'katello_devel::setup':
-      require   => Class['katello::candlepin'],
+      require   => [Class['katello::candlepin'], Class['katello::pulp']],
```
ekohl: Why is this needed?

manifests/init.pp
```diff
@@ -178,37 +178,11 @@
...
   }
 
-  User<|title == $user|>{groups +> 'qpidd'}
-
-  # TODO: Use katello::pulp
-  class { 'certs::qpid_client':
-    require => Class['pulp::install'],
-    notify  => Class['pulp::service'],
+  class { 'katello::pulp':
+    require => Class['katello_devel::database']
```
ekohl: Pulp (2) doesn't use the postgresql database, so why is this needed?

ehelms:
All cleaned up and ready to party

manifests/init.pp
```diff
@@ -178,37 +178,9 @@
     require => Class['katello_devel::database'],
   }
 
-  User<|title == $user|>{groups +> 'qpidd'}
-
-  # TODO: Use katello::pulp
```
ekohl: Yay for cleaning up my TODOs

PR #226, opened 2020-02-17
Changed Puma default workers to 2

sthirugn:
To sync with the default Puma settings in https://github.com/theforeman/foreman/pull/7421

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -49,7 +49,7 @@ class katello_devel::params {
 
   $qpid_wcache_page_size = 4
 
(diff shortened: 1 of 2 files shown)


ekohl:
Looks like the tests need an update as well

sthirugn:
@ekohl spec tests updated.

ekohl:
For future reference, we typically don't create Redmine issues for the devel module since it has no release cycle. There's only git master.

sthirugn:
> For future reference, we typically don't create Redmine issues for the devel module since it has no release cycle. There's only git master.

Great thank you for the info.

PR #227, opened 2020-02-26
Connect to candlepin via localhost

jturel:


ekohl:
Alternative implementation at https://github.com/theforeman/puppet-katello_devel/pull/228 which should work with either version (as long as the params keep exposing the URL.

PR #230, opened 2020-03-06
Add postgresql-evr extension

ianballou:
Adds the postgresql-evr extension needed for the upcoming `evt_t` postgresql type.  This won't work out of the box until the package is built in release but it is in staging currently.


jlsherrill:
ACK pending nightly passing (and that package appearing there)

PR #245, opened 2020-11-03
Support CentOS 8 and Pulp 3 only installs

ehelms:


manifests/params.pp
```diff
@@ -21,20 +21,25 @@
   $use_scl_ruby = true
   $scl_ruby = 'rh-ruby25'
 
-  $scl_nodejs = 'rh-nodejs12'
-  $scl_postgresql = 'rh-postgresql12'
+  if $facts['os']['release']['major'] == '7' {
+    $scl_nodejs = 'rh-nodejs12-'
+    $scl_postgresql = 'rh-postgresql12-'
```
ekohl, line 26: This is a bit inconsistent with the above `$scl_ruby` which doesn't use the `-` suffix which is a bit annoying.

manifests/params.pp
```diff
@@ -21,20 +21,25 @@
   $use_scl_ruby = true
   $scl_ruby = 'rh-ruby25'
```
ekohl, line 22: Shouldn't this also be only on EL7?

manifests/params.pp
```diff
@@ -21,20 +21,25 @@
...
+    $scl_postgresql = undef
+  }
 
   $manage_bundler = true
 
   $initial_organization = 'Default Organization'
   $initial_location = 'Default Location'
   $admin_password = 'changeme'
-  $enable_ostree = true
+  $enable_ostree = $facts['os']['release']['major'] == '7'
```
ekohl, line 37: Perhaps it's better to include `katello::params` and reuse `$katello::params::pulp2_support` for this?

ekohl (changes requested):
I opened https://github.com/theforeman/puppet-katello_devel/pull/246 which should make it easier to deal with optional SCLs. I'd prefer to base it on that.

manifests/params.pp
ehelms: A bit, but these are are also not inputs from the user (which none of them should be).

manifests/params.pp
ehelms: Theoretically, we allow users to pass this in for some reason so I left it alone.

manifests/params.pp
ekohl: Still, I strongly dislike inconsistency. This also means you can't enable it by default. For example, in forklift we pass the Ruby version here:
https://github.com/theforeman/forklift/blob/b85b14b7f894b9ee1193fa15f8780d887f13388c/roles/katello_devel/meta/main.yml#L18

It can't have the dash since the variable is also used here:
https://github.com/theforeman/forklift/blob/b85b14b7f894b9ee1193fa15f8780d887f13388c/roles/ruby_scl/tasks/main.yml#L14-L22

Now we do the same thing here:
https://github.com/theforeman/forklift/blob/b85b14b7f894b9ee1193fa15f8780d887f13388c/roles/nodejs_scl/tasks/main.yml#L12-L18

That's why I think they must be parameters. That way they remain in sync.

The alternative is to kill the nodejs_scl and ruby_scl Ansible roles and do it all in this module.

manifests/install.pp
```diff
@@ -6,28 +6,25 @@
...
       'gcc-c++',
       'libstdc++',
-      "${katello_devel::scl_postgresql}-postgresql-debversion",
-      "${katello_devel::scl_postgresql}-postgresql-evr",
+      katello_devel::package('postgresql-debversion', $katello_devel::scl_postgresql),
+      katello_devel::package('postgresql-evr', $katello_devel::scl_postgresql),
       'katello-selinux',
+      'make',
+      katello_devel::package('ruby-devel', $katello_devel::scl_ruby),
+      katello_devel::package('rubygem-bundler', $katello_devel::scl_ruby),
```
ekohl, line 23: If you do this, you perhaps the RVM support should be dropped? I wouldn't mid that at all btw but I don't know if anyone is still using it.

functions/package.pp
```diff
@@ -0,0 +1,21 @@
...
+#   An optional SCL name, without a dash as suffix
+#
+# @return The package name, with SCL prefix if applicable
+#
+# @example Without SCL
+#   katello_devel::package('postgresql-server') == ['postgresql-server']
+#
+# @example With SCL
+#   katello_devel::package('postgresql-server', 'rh-postgresql12') == ['rh-postgresql12-postgresql-server']
+function katello_devel::package(String[1] $package, Optional[String[1]] $scl = undef) >> String[1] {
```
ekohl, line 15: Would you mind copying (+ changing) the tests I wrote in the other PR as well?

ehelms: I did -- https://github.com/theforeman/puppet-katello_devel/pull/245/files#diff-88cf9c8f9f8eb64329bd90c44afc454a89dfa0acaa823380fcc8a5a3b6751164

manifests/install.pp
ehelms: I have been considering that, just didn't want to muddy the waters of this PR with that change. It benefits devel envs to mirror production even at how Ruby is provided level.

spec/functions/packages_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+describe 'katello_devel::package' do
+  describe 'without scl' do
+    it { is_expected.to run.with_params().and_return() }
+    it { is_expected.to run.with_params('foo').and_return(['foo']) }
```
ekohl, line 6:
```suggestion
    it { is_expected.to run.with_params('foo').and_return('foo') }
```

spec/functions/packages_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+describe 'katello_devel::package' do
+  describe 'without scl' do
+    it { is_expected.to run.with_params().and_return() }
```
ekohl, line 5: I'd expect this to error out since it requires a non-empty string as first argument.

manifests/params.pp
ehelms: That threw all kinds of duplicate declaration errors so I just opted to stay with the easy path.

manifests/init.pp
```diff
@@ -100,7 +100,7 @@
   String $rvm_ruby = $katello_devel::params::rvm_ruby,
   String $rvm_branch = $katello_devel::params::rvm_branch,
   Boolean $use_scl_ruby = $katello_devel::params::use_scl_ruby,
```
ekohl, line 102: I do wonder what's the use of this if `$scl_ruby` can be undef. Perhaps it should be dropped instead by changing this part:
https://github.com/theforeman/puppet-katello_devel/blob/230b6165953b079c84c4396f86269fea60db8e7b/manifests/bundle.pp#L22
Make it `if $scl_ruby` instead.

ekohl (approved):

ehelms:
I dropped `use_scl_ruby` which will break the current Forklift role, so I will time this with merging https://github.com/theforeman/forklift/pull/1218 after I run an additional test.

PR #246, opened 2020-11-04
Add the scl_packages function

ekohl:
This makes it easier to make the SCL variables optional, which is useful for EL8.

### DIFF
--- a/functions/scl_packages.pp
+++ b/functions/scl_packages.pp
@@ -1,0 +1,19 @@
+# @summary Generate package names that may or may not be in some SCL
+# @param packages
+#   The package names without any SCL prefix
+# @param scl
+#   An optional SCL name, without a dash as suffix
+# @return The package names, with SCL prefix if applicable
+#
+# @example Without SCL
+#   katello_devel::scl_packages(['postgresql-server']) == ['postgresql-server']
+#
+# @example With SCL
+#   katello_devel::scl_packages(['postgresql-server'], 'rh-postgresql12') == ['rh-postgresql12-postgresql-server']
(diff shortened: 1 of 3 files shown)


ehelms:
Can you expand on why this makes it easier?

ekohl:
Now you can set `$scl_x = undef` in `params.pp` and it'll still generate the right packages. This allows you to be consistent.

functions/scl_packages.pp
```diff
@@ -0,0 +1,19 @@
...
+# @param scl
+#   An optional SCL name, without a dash as suffix
+# @return The package names, with SCL prefix if applicable
+#
+# @example Without SCL
+#   katello_devel::scl_packages(['postgresql-server']) == ['postgresql-server']
+#
+# @example With SCL
+#   katello_devel::scl_packages(['postgresql-server'], 'rh-postgresql12') == ['rh-postgresql12-postgresql-server']
+function katello_devel::scl_packages(Array[String[1]] $packages, Optional[String[1]] $scl = undef) >> Array[String[1]] {
```
ehelms, line 13: Since this can return non-SCL packages, I think the function should be called `packages` as it gives a false impression when you read it's use in code. Something like (https://github.com/theforeman/puppet-katello_devel/pull/245/files#diff-29fdc8c170b6278c50ae7a6359737e386dc4769ce94b9a6e505f5bcd3e4890deR26-R34) is repetitive in a sense, but it is also incredibly clear and easy to read what is happening for a developer.

ekohl:
Closing since you're integrating the same thing in the other PR. I just thought this was the easiest way to show the idea.

PR #247, opened 2020-12-02
Properly deploy pulp2 config

jlsherrill:


jlsherrill:
@ehelms i have no idea if this is the right way to do it, but this seems to work.   The previous @pulp2_support was never set to true

manifests/config.pp
```diff
@@ -5,10 +5,9 @@
   $user = $katello_devel::user,
   $group = $katello_devel::group,
   $extra_plugins = $katello_devel::extra_plugins,
+  $pulp2_support = $katello::params::pulp2_support,
```
ekohl, line 8: This change shouldn't be needed but doesn't hurt either. If you want to be stricter, you can ensure that it's set. That way compilation fails if you would depend on undefined behavior.
```suggestion
  Boolean $pulp2_support = $katello::params::pulp2_support,
```

ekohl (approved):
Small suggestion inline but overall :+1:

PR #248, opened 2020-12-08
support passing katello and foreman branches/revisions

jlsherrill:


jlsherrill:
This is untested yet, but opening for early looks.   The goal is to generate vagrant boxes for each katello release (so we have a fully functioning dev box months after release time).  This should allow the installer to be passed versions to checkout foreman and katello versions for a given release.

jlsherrill:
This is now tested and works as intended

ekohl (approved):
I think it makes sense. At some point the `extra_plugins` parameter may be desired as well. I can imagine the REX plugin needs to be some older branch as well, but it'd be fine to do so in a future version when the need arises.

jlsherrill:
thanks @ekohl !

PR #257, opened 2021-04-22
Don't add vagrant to qpidd group

jturel:


ekohl (approved):
I was recently looking at this exact line and thought it could be dropped.

jturel:
> I was recently looking at this exact line and thought it could be dropped.

Yeah it wasn't even needed with the agent infrastructure is there :)

PR #259, opened 2021-05-07
Convert parameters to static defaults

ehelms:


manifests/init.pp
```diff
@@ -90,35 +90,35 @@
...
+  Boolean $enable_file = true,
+  Boolean $enable_docker = true,
+  Boolean $enable_deb = true,
+  Boolean $enable_ansible_collection = true,
+  Optional[String] $github_username = undef,
+  Boolean $use_ssh_fork = false,
+  Optional[String] $fork_remote_name = undef,
+  String $upstream_remote_name = 'upstream',
+  Integer[0, 1000] $qpid_wcache_page_size = 4,
+  Array[String] $extra_plugins = [],
```
ekohl, line 117: We've seen in the past that array defaults fail.

ekohl (changes requested):

PR #261, opened 2021-05-18
[CP] 3.18 cherry-picks

jturel:
the 3.18 branch was cut from a commit a bit too early - this brings over what's missing and will likely fix the broken packer build


manifests/params.pp
```diff
@@ -41,6 +41,9 @@
   $fork_remote_name = undef
   $upstream_remote_name = 'upstream'
 
+  $foreman_scm_revision = 'develop'
+  $katello_scm_revision = 'master'
```
ekohl, line 45: Shouldn't these be 3.18 and whatever Foreman matches that? (2.3?) That way the defaults are correct if you check out the branch.

jeremylenz, line 45: I wonder if there's a list of places we need to change when this changes to `main`..

jturel: They get overridden by our packer scripting via installer flags, I don't think anyone is directly checking these branches out ... yet

jeremylenz (approved):
LGTM

manifests/params.pp
jturel: Nonexistent afaik

jturel:
@ekohl any idea what's up with the unit tests here?

metadata.json
```diff
@@ -30,7 +30,7 @@
     },
     {
       "name": "katello-katello",
-      "version_requirement": ">= 6.1.0"
+      "version_requirement": "<= 16.0.0"
```
jturel, line 33: @ekohl does this pinning not affect the dependencies used by the unit tests here?

ekohl: No, that's in `.fixtures.yml`.

jturel:
Getting somewhere. need to pin more things

PR #264, opened 2021-07-19
Drop foreman_certs class and let puppet-certs control

ehelms:
Results from this change https://github.com/theforeman/puppet-certs/pull/371


ehelms:
Puppet 5 failure would be mitigated by dropping it https://github.com/theforeman/puppet-katello_devel/pull/262

manifests/foreman_certs.pp
```diff
@@ -1,45 +0,0 @@
...
-
-  Cert[$client_cert_name] ~>
-  pubkey { $client_cert:
-    key_pair => Cert[$client_cert_name],
-  } ~>
-  privkey { $client_key:
-    key_pair => Cert[$client_cert_name],
-  } ~>
-  file { $client_key:
-    group => $katello_devel::group,
```
ekohl, line 40: This will break, right? The group is different. What I think we should do is reuse [certs::foreman](https://github.com/theforeman/puppet-certs/blob/master/manifests/foreman.pp) but update it to accept a group parameter. Then we can use it like this:
```puppet
class { 'certs::foreman':
  group => $katello_devel::group,
}
```

Remember why I never want hardcoded values but always parameters, even if we don't immediately use them? this is why :)

ekohl (changes requested):

manifests/foreman_certs.pp
jturel: Seems like a pretty quick change to make. Waiting for resolution (one way or the other) before I test

ehelms: Wellll, there are a couple nuances to it we could go by here. So let me layout the current, and sorta two approaches.

##### Current

The development environment deploys the following for Foreman client certificates on a standard Vagrant environment:

```
/etc/pki/katello/certs/foreman.example.com-foreman-client.crt
/etc/pki/katello/private/foreman.example.com-foreman-client.key
```

Where the certificate is world readable and the key is `root:vagrant` with `440` permissions.

A production installation deploys as:

```
/etc/foreman/client_cert.pem
/etc/foreman/client_key.pem
```

With `root:foreman` ownership for both and `440` for both.

##### Solutions

Reminder, these are only for development environments so trying to balance matching production deployments as closely as possible with ease of development environments and discoverability.

Option 1

Allow a production like deployment of client certificates to `/etc/foreman` and set permissions to `root:vagrant`

Option 2

Allow a production like deployment of client certificates to `/etc/foreman` and put the vagrant user in the foreman group.

Option 3

Deploy client certificates to `/home/vagrant/foreman/config` with `vagrant:vagrant` ownership.

jturel: I like (2) but would be fine with just (1) for now. Guessing there's a bit more work to be done in order to achieve (2)?

ehelms: I *think* 2 is easier because you just remove all this code, and use a virtual resource to add the user to the foreman group.

jturel: Sounds good to me. Let's see what @ekohl thinks

evgeni: +1 on the user in the group

ehelms: Option 3 would require some updates to `.gitignore` in Foreman to specifically ignore the files we are laying down or a special directory inside `config/` to put temporary devel environment files (e.g. `config/certs` or `config/tls`). This could be nice from a more encapsulated deployment perspective.

ekohl: You can of course also place them in `/home/vagrant/foreman-ssl` and point to them. Perhaps that's even better since then users can wipe their git checkout without cleaning the certs.

ehelms: I like the idea of putting the certs in a more dev friendly location longer term. In the short term, to re-use production code and reduce scope in here, I think we'll need the owner/group setting (here https://github.com/theforeman/puppet-certs/pull/374) and then have this declare that and create `/etc/foreman`.

manifests/init.pp
```diff
@@ -162,11 +162,22 @@
 
   include certs::pulp_client
 
+  file { '/etc/foreman':
```
ekohl, line 165: I still feel placing the certs in `$deployment_dir` or `${deployment_dir}/certs` is cleaner than `/etc/foreman`.

manifests/init.pp
```diff
@@ -162,11 +162,22 @@
...
+    ensure => directory,
+    owner  => 'root',
+    group  => $group,
+    mode   => '0775'
+  }
+
+  class { 'certs::foreman':
```
ekohl, line 172: Please make sure this is executed before `katello_devel::config`

manifests/init.pp
```diff
@@ -162,11 +162,22 @@
 
   include certs::pulp_client
```
ekohl, line 163: Unrelated, but do we still need this?

ekohl (changes requested):
You'll need to modify this too:
https://github.com/theforeman/puppet-katello_devel/blob/13463e9ad0174150f974020e21bf84a5c078f009/templates/settings.yaml.erb#L16-L19

manifests/init.pp
ehelms: In theory yes, for N-1 testing in devel environments this could be useful.

manifests/foreman_certs.pp
ekohl: You already have the parameters (`client_cert`, `client_key`, and `ssl_ca_cert`) so you can easily modify those. I'm not sure it really increases the scope too much.

ehelms: Because of the way https://github.com/theforeman/puppet-certs/blob/master/manifests/foreman.pp works, I would still have to ensure `/etc/foreman` exists, or set deploy to false and then implement something akin to https://github.com/theforeman/puppet-certs/pull/370. So my general point is, right now, to unblock devel its easier to abide by the production deployment model and then as we work through making puppet-certs more flexible, we consider adjusting as part of that how we deploy the devel env.

ekohl: Is that because the `cert` resource itself is putting something in `/etc/foreman`? Because I don't really see what else requires it if you pass the parameters.

ehelms: `privkey` and `pubkey` handle putting things into `/etc/foreman`. However, if I pass `deploy => false` to that resource, then the RPMs won't get installed that place the certificate and key into `/etc/pki/katello-certs-tools` which is what `privkey` and `pubkey` work based upon. So, that is why I need the solution of deploying from the build dir (https://github.com/theforeman/puppet-certs/pull/370) which I could use puppet-katello_devel to pilot that.

I want to unblock developers first, and then look to make changes to the environment. Thus I would like to go with this solution and follow up with the build directory deployment model.

ekohl (approved):

manifests/foreman_certs.pp
ehelms: Full circle -- https://github.com/theforeman/puppet-katello_devel/issues/265

ekohl: For what it's worth: I ran https://github.com/ekohl/puppet-certs/commit/adec2d6dfb74a86bf76dfdd2f49d607f895817be locally and it passes. No `/etc/foreman` exists. IMHO there is no reason to create `/etc/foreman`. While I didn't test https://github.com/ekohl/puppet-katello_devel/commit/8e2f945cd4e70d8e39737ad5a90c3e431442f22b, based on the puppet-certs test it should work.

ehelms: That should in theory work, I'll give it a go, thanks!

jturel:
Testing this, I think it's good but I'm seeing unrelated errors.

1 - had a problem with seeding remote execution job templates, worked around that
2 - I think is related to the puppet removal?

```
    2021-07-21 18:43:20 [ERROR ] [configure] Proxy centos7-katello-devel.strangeways.example.com has failed to load one or more features (Puppet), check /var/log/foreman-proxy/proxy.log for configuration errors
    2021-07-21 18:43:20 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-katello-devel.strangeways.example.com]/ensure: change from 'absent' to 'present' failed: Proxy centos7-katello-devel.strangeways.example.com has failed to load one or more features (Puppet), check /var/log/foreman-proxy/proxy.log for configuration errors
    2021-07-21 18:43:21 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-katello-devel.strangeways.example.com]: Failed to call refresh: Proxy centos7-katello-devel.strangeways.example.com has failed to load one or more features (Puppet), check /var/log/foreman-proxy/proxy.log for configuration errors
    2021-07-21 18:43:21 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-katello-devel.strangeways.example.com]: Proxy centos7-katello-devel.strangeways.example.com has failed to load one or more features (Puppet), check /var/log/foreman-proxy/proxy.log for configuration errors
    2021-07-21 18:43:23 [NOTICE] [configure] System configuration has finished.

```

Not seeing any errors in proxy.log

```
2021-07-21T18:43:11  [I] WEBrick::HTTPServer#start done.
2021-07-21T18:43:13  [W] Could not open DB for dynflow at '', will keep data in memory. Restart will drop all dynflow data.
2021-07-21T18:43:13  [I] Execution plan cleaner removing 0 execution plans.
2021-07-21T18:43:13  [I] Successfully initialized 'dynflow'
2021-07-21T18:43:13  [I] Successfully initialized 'ssh'
2021-07-21T18:43:13  [I] Successfully initialized 'pulpcore'
2021-07-21T18:43:13  [I] Successfully initialized 'foreman_proxy'
2021-07-21T18:43:13  [I] Successfully initialized 'puppetca_http_api'
2021-07-21T18:43:13  [I] Successfully initialized 'puppetca_hostname_whitelisting'
2021-07-21T18:43:13  [I] Successfully initialized 'puppetca'
2021-07-21T18:43:13  [I] Successfully initialized 'puppet_proxy_puppet_api'
2021-07-21T18:43:13  [I] Successfully initialized 'puppet'
2021-07-21T18:43:13  [I] Successfully initialized 'logs'
2021-07-21T18:43:13  [I] Successfully initialized 'registration'
2021-07-21T18:43:13  [I] WEBrick 1.6.1
2021-07-21T18:43:13  [I] ruby 2.7.3 (2021-04-05) [x86_64-linux]
2021-07-21T18:43:13  [I]
```

ekohl:
>  2 - I think is related to the puppet removal?

Yes. The Puppet module is enabled but on the Foreman side nothing recognizes it. The installer doesn't like that because it indicates a missing plugin. The solution is to either install the puppet plugin but what's better is to disable the puppet. I'll see if I can come up with a PR.

ekohl:
If you can give https://github.com/theforeman/forklift/pull/1392 a spin, that would be great.

jturel:
> If you can give [theforeman/forklift#1392](https://github.com/theforeman/forklift/pull/1392) a spin, that would be great.

Started with a fresh VM and now I see this which is new:

```
    foreman-installer -v --no-colors --disable-system-checks --scenario katello-devel  --katello-devel-scl-ruby=rh-ruby27 --katello-devel-admin-password changeme --katello-devel-github-username=jturel --katello-devel-extra-plugins theforeman/foreman_remote_execution
  delta: '0:01:39.964667'
  end: '2021-07-21 19:11:04.397388'
  msg: non-zero return code
  rc: 1
  start: '2021-07-21 19:09:24.432721'
  stderr: ''
  stderr_lines: <omitted>
  stdout: |-
    2021-07-21 19:09:29 [NOTICE] [root] Loading installer configuration. This will take some time.
    2021-07-21 19:09:32 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
    2021-07-21 19:09:32 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
    2021-07-21 19:09:33 [WARN  ] [pre] Skipping system checks.
    2021-07-21 19:09:33 [WARN  ] [pre] Skipping system checks.
    2021-07-21 19:10:56 [NOTICE] [configure] Starting system configuration.
    2021-07-21 19:11:04 [ERROR ] [configure] Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: File[/etc/foreman-proxy/ssl_key.pem] is already declared at (file: /usr/share/foreman-installer/modules/foreman_proxy/manifests/config.pp, line: 135); cannot redeclare (file: /usr/share/foreman-installer/modules/certs/manifests/keypair.pp, line: 36) (file: /usr/share/foreman-installer/modules/certs/manifests/keypair.pp, line: 36, column: 5) (file: /usr/share/foreman-installer/modules/certs/manifests/foreman_proxy.pp, line: 95) on node centos7-katello-devel.strangeways.example.com
    2021-07-21 19:11:04 [NOTICE] [configure] System configuration has finished.
  
      There were errors detected during install.
      Please address the errors and re-run the installer to ensure the system is properly configured.
      Failing to do so is likely to result in broken functionality.
  
      The full log is at /var/log/foreman-installer/katello-devel.log
  stdout_lines: <omitted>
```

ehelms:
I went ahead and implemented the `foreman-certs/` change after seeing how it could be possible from @ekohl . I think we'll need https://github.com/theforeman/puppet-certs/pull/375 present.

ehelms:
@jturel try adding https://github.com/theforeman/forklift/pull/1392/files#r674270152 locally

jturel:
> @jturel try adding https://github.com/theforeman/forklift/pull/1392/files#r674270152 locally

That did the trick!

jturel:
This looks good:

```
  :candlepin_events:
    :ssl_cert_file: /home/vagrant/foreman-certs/client_cert.pem
    :ssl_key_file: /home/vagrant/foreman-certs/client_key.pem
    :ssl_ca_file: /etc/pki/katello/certs/katello-default-ca.crt
```

```
[vagrant@centos7-katello-devel ~]$ ll foreman-certs/
total 20
-r--r-----. 1 vagrant vagrant 8417 Jul 21 19:57 client_cert.pem
-r--r-----. 1 vagrant vagrant 3243 Jul 21 19:57 client_key.pem
-r--r-----. 1 vagrant vagrant 2549 Jul 21 19:57 proxy_ca.pem
```

```
                "candlepin_events": {
                    "duration_ms": "0",
                    "message": "0 Processed, 0 Failed",
                    "status": "ok"
                },
```

jturel (approved):

manifests/init.pp
```diff
@@ -162,11 +163,30 @@
...
+    client_key  => "${foreman_cert_dir}/client_key.pem",
+    ssl_ca_cert => "${foreman_cert_dir}/proxy_ca.pem",
+    owner       => $user,
+    group       => $group,
+    before      => Class['katello_devel::config'],
+  }
+
+  $ssl_ca_file = "${foreman_cert_dir}/proxy_ca.pem"
+  $ssl_certificate = "${foreman_cert_dir}/client_cert.pem"
+  $ssl_priv_key = "${foreman_cert_dir}/client_key.pem"
```
ekohl, line 184: If you define it like this, then I'd say it's better to define it earlier and use the variables as parameters to `certs::foreman`: https://github.com/theforeman/puppet-katello_devel/pull/267

PR #266, opened 2021-07-21
Add basic acceptance test

ehelms:
This is *very* basic at the moment in terms of what it verifies. The biggest win I am aiming for here is that PRs will get automated testing of the module and help speed up fixes for developers.


ehelms:
@jturel you might be interested in this as it could speed up test cycles

jturel:
I like it. Seems like the new tests weren't run based on the github action logs, or maybe I missed it?

spec/setup_acceptance_node.pp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+# Necessary for PostgreSQL EVR extension
+yumrepo { "pulpcore":
+  baseurl  => "http://yum.theforeman.org/pulpcore/3.14/el\$releasever/x86_64/",
+  descr    => "Pulpcore",
+  enabled  => true,
+  gpgcheck => true,
+  gpgkey   => "https://yum.theforeman.org/pulpcore/3.14/GPG-RPM-KEY-pulpcore",
+}
```
ekohl, line 23: Not just for EVR, also Pulp itself. However, we should have the Pulpcore module here
```suggestion
class { 'pulpcore::repo':
}
```

ehelms:
> I like it. Seems like the new tests weren't run based on the github action logs, or maybe I missed it?

Good catch, I forget that our modulesync only adds acceptance tests to the github workflow if there are actual acceptance tests. I've added that now.

spec/setup_acceptance_node.pp
ehelms: We should, if this module actually did anything with Pulpcore :) This is similar to the puppet-katello acceptance tests, this does not rely on puppet-foreman_proxy_content which handles setting up Pulp 3. And for the installer, the answer file brings the modules together. This module has no knowledge of getting Pulp set up.

ekohl: Ah right

ekohl (approved):

PR #269, opened 2021-10-08
install centos-qpid repo before installing qpid-devel on el8

evgeni:


evgeni:
mind you, I didn't have time to test it yet

ehelms:
Knowing is half the battle

PR #270, opened 2021-11-08
remove content types

jlsherrill:


jlsherrill:
looks like some of the recent certs changes are breaking this more

 Unknown variable: 'certs::apache::apache_ca_cert'.

manifests/apache.pp
```diff
@@ -6,8 +6,8 @@
     ssl           => true,
     ssl_cert      => $certs::apache::apache_cert,
     ssl_key       => $certs::apache::apache_key,
-    ssl_ca        => $certs::apache::ca_cert,
-    ssl_chain     => $certs::apache::apache_ca_cert,
+    ssl_ca        => $certs::katello_server_ca_cert,
+    ssl_chain     => $certs::katello::server_ca_cert,
```
ehelms, line 10:
```suggestion
    ssl_chain     => $certs::katello_server_ca_cert,
```

manifests/apache.pp
```diff
@@ -6,8 +6,8 @@
     ssl           => true,
     ssl_cert      => $certs::apache::apache_cert,
     ssl_key       => $certs::apache::apache_key,
-    ssl_ca        => $certs::apache::ca_cert,
-    ssl_chain     => $certs::apache::apache_ca_cert,
```
ekohl, line 10: I think that removing these variables is actually a regression. IMHO you as a consumer of `certs::apache` don't need to think about the actual CA: it's just there as a variable. Can't we bring that back @ehelms?

ehelms: It was removed because it was unused within the `certs::apache` class and therefore existed "internally" and entirely unused within the code itself. I think that's a bit of the tricky part -- is a variable within a class an external API? I recall we've had some fun debates on this.

ekohl: I considered it part of the external API, precisely for this use case.

ekohl: Also note that this change is breaking. `ca_cert` is the default CA (ssl_ca, used to authenticate with client certs) where the chain is the actual CA that signed `$apache_cert`.

I've opened https://github.com/theforeman/puppet-certs/pull/384.

ekohl: Yes, that looks like an oversight: https://github.com/theforeman/puppet-foreman_proxy_content/pull/392

ehelms: @jlsherrill you can drop this update now and just have the content types

ehelms:
Unit tests will need updates

ehelms (approved):

PR #273, opened 2022-04-28
include apache::mod::setenvif

wbclark:
puppet-katello included this as of
    https://github.com/theforeman/puppet-katello/commit/afc08e300edb831d8821cb335a341afbca161cf1

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -11,6 +11,9 @@ class katello_devel::apache {
     proxy_backend => "http://localhost:${katello_devel::rails_port}/",
   }
 
+  # required by configuration in katello/katello-apache-ssl.conf
+  include apache::mod::setenvif
+
   foreman::config::apache::fragment { 'katello':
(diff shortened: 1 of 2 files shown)


wbclark:
This is needed due to https://github.com/theforeman/foreman-installer/pull/754 and fixed in katello with https://github.com/theforeman/puppet-katello/pull/445

ehelms (approved):

sjha4:
Not to hijack this PR but putting it here since we have infrastructure gurus on the thread. There's another issue in the repo that is breaking dev boxes:

The package 'cyrus-sasl-plain' is declared twice for dev box:

1. https://github.com/theforeman/puppet-katello_devel/blob/master/manifests/install.pp
2. https://github.com/theforeman/puppet-qpid/blob/master/manifests/install.pp#L21

This throws a duplicate declaration error on vagrant up of dev boxes.

sjha4:
Is the test failure related? We should merge this if not.

PR #274, opened 2022-05-04
Ensure setenvif module is active

poelzi:
Using setenvif directive without enableing module broke forklift and seems wrong


sjha4:
Should be fixed by this: [theforeman/puppet-katello_devel#273](https://github.com/theforeman/puppet-katello_devel/pull/273) ?

PR #277, opened 2022-05-19
Allow setting nodejs module stream and default to 12 for EL8+

wbclark:


manifests/init.pp
```diff
@@ -85,6 +88,7 @@
   String $rvm_ruby = '2.7',
   String $rvm_branch = 'stable',
   Optional[String] $scl_ruby = $katello_devel::params::scl_ruby,
+  Optional[String] $modulestream_nodejs = $katello_devel::params::modulestream_nodejs,
```
wbclark, line 91: I've tested and confirmed this works on EL8 with node version 12 and 14. Therefore I should probably change this type to an `Enum` when I have some time later.

BTW, regarding the `npm install` timeout observed in CI, on my local machine it took ~32 minutes in each test case. The default value probably needs to be updated here as well.

wbclark, line 91: I've tested and confirmed this works on EL8 with node version 12 and 14. Therefore I should probably change this type to an `Enum` when I have some time later.

BTW, regarding the `npm install` timeout observed in CI, on my local machine it took ~32 minutes in each test case. The default value probably needs to be updated here as well.

chris1984 (approved):
ACK, looks fine. @wbclark can you get the other tests passing?

wbclark:
Re-running the failed tests, let's see

wbclark:
"Failed to download metadata for repo 'foreman': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried"

PR #278, opened 2022-05-23
Increase npm_timeout default to 45 minutes

wbclark:
Previously this was set to 900 seconds (15 minutes). More recently,
the `npm install` step has been observed to take over 30 minutes.
This was causing failures in local installation as well as in the
CI environment.


ehelms (approved):

ehelms:
I'll merge to unblock, but we should really look into making that less of an issue. 45 minutes?!

PR #280, opened 2022-06-08
Drop EL7 support

ehelms:


ehelms:
This made me wonder if we should (in a separate follow up) drop the use of RVM all together and rely solely on available modules for development environments.

manifests/init.pp
```diff
@@ -87,8 +84,7 @@
   Boolean $use_rvm = false,
   String $rvm_ruby = '2.7',
   String $rvm_branch = 'stable',
-  Optional[String] $scl_ruby = $katello_devel::params::scl_ruby,
-  Optional[String] $modulestream_nodejs = $katello_devel::params::modulestream_nodejs,
+  Optional[String] $modulestream_nodejs = '12',
```
ekohl, line 87: If you provide a default, you shouldn't have `Optional`
```suggestion
  String $modulestream_nodejs = '12',
```

manifests/install.pp
```diff
@@ -4,31 +4,29 @@
 
   $modulestream_nodejs = $katello_devel::modulestream_nodejs
 
-  if $modulestream_nodejs != undef {
-    package { 'nodejs':
-      ensure   => $modulestream_nodejs,
-      provider => 'dnfmodule',
-    }
+  package { 'nodejs':
```
ekohl, line 7: Looking at this, it always should have had a `before => Package['npm']`. Would you mind adding it while you're at it?

manifests/install.pp
```diff
@@ -4,31 +4,29 @@
 
   $modulestream_nodejs = $katello_devel::modulestream_nodejs
```
ekohl, line 5: You can consider removing this variable and in the `ensure` use `$katello_devel::modulestream_nodejs`.

ekohl (commented):
My changes did also create a merge conflict. Other than that :+1: 

I've thought about dropping RVM too and I think it's a good idea. I even played around with writing patches for it. Looks like I did push a branch but forgot to open the PR for it. I rebased it and opened https://github.com/theforeman/puppet-katello_devel/pull/282. It's only just over a year old ;)

PR #282, opened 2022-06-22
Drop RVM support

ekohl:
It already defaulted to OS versions since that's also used in production.


ekohl:
IMHO we should create a `modulestream_ruby` parameter for that, just like we have a `modulestream_nodejs` parameter.

ehelms:
> IMHO we should create a `modulestream_ruby` parameter for that, just like we have a `modulestream_nodejs` parameter.

I think my point was missed. RVM allows co-installation of Ruby versions that is faster to move between when doing development work that targets new versions of Ruby where things are different between the versions (e.g. 2.7 vs. 3.0). The module method does not make this easy for developers.

ehelms:
Two development machines can be prohibitive for users and makes testing back and forth between the two Ruby versions difficult. I don't think the installer in this case will play a role as the development environment does not really support being re-run. I would expect that developers wanting RVM setup an environment and then manage installing additional Ruby and gemsets and swapping their setup between them themselves. Providing RVM setup here makes the initial setup and configuration as an option easy for developers.

ekohl:
I think if users want RVM, they can follow what's on https://rvm.io/ instead of us trying to provide something that isn't really supported.

ehelms:
I'd like to get a few developers to weigh in -- @wbclark @chris1984 perhaps you could give your thoughts?

ehelms (approved):

PR #283, opened 2022-08-01
Add an option to use the user's fork as the default remote

wbclark:


wbclark:
Test failures look related to https://github.com/theforeman/puppet-foreman/pull/1070

I'll try to pin puppetlabs-apache version in .fixtures.yml for now

.fixtures.yml
```diff
@@ -2,7 +2,9 @@ fixtures:
   repositories:
     stdlib:        "https://github.com/puppetlabs/puppetlabs-stdlib.git"
     extlib:        "https://github.com/voxpupuli/puppet-extlib"
-    apache:        "https://github.com/puppetlabs/puppetlabs-apache.git"
+    apache:
+      repo: 'https://github.com/puppetlabs/puppetlabs-apache.git'
+      ref: 'v7.0.0'
```
ekohl, line 7: The fix was merged. I think you can revert this again.

wbclark: Reverted, TY

wbclark:
@ekohl looks like we're now facing `Apache::Vhost[foreman-ssl]: parameter 'ssl_verify_depth' expects a value of type Undef or Integer, got String`. Likely explanation is the same commit in puppetlabs-apache -- would you consider that a regression as well?

wbclark:
Indeed, I found the change which caused it.

wbclark:
Opened https://github.com/theforeman/puppet-foreman/pull/1071

chris1984:
Code looks good, tested this and got this error:

```bash
[vagrant@centos82-katello-devel ~]$ sudo foreman-installer --scenario katello-devel --katello-devel-fork-as-default-remote true --katello-devel-github-username chris1984
2022-08-04 20:02:38 [NOTICE] [root] Loading installer configuration. This will take some time.
2022-08-04 20:02:40 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
2022-08-04 20:02:40 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
2022-08-04 20:02:43 [NOTICE] [configure] Starting system configuration.
2022-08-04 20:02:50 [NOTICE] [configure] 250 configuration steps out of 1468 steps complete.
2022-08-04 20:02:54 [NOTICE] [configure] 500 configuration steps out of 1473 steps complete.
2022-08-04 20:02:55 [NOTICE] [configure] 750 configuration steps out of 1477 steps complete.
2022-08-04 20:02:55 [NOTICE] [configure] 1000 configuration steps out of 1481 steps complete.
2022-08-04 20:02:57 [ERROR ] [configure] /Stage[main]/Katello_devel::Install/Katello_devel::Git_repo[foreman]/Vcsrepo[/home/vagrant/foreman]: Could not evaluate: Execution of 'git fetch chris1984' returned 128: fatal: could not read Password for 'https://chris1984@github.com': No such device or address
```

ekohl (commented):
I'm confused by this. What do you mean as default remote?

For background: git doesn't really have a default remote. Here's how I usually work with it:
* I clone upstream as upstream: `git clone -o upstream UPSTREAM_URL`
* I add a remote with my username: `git remote add ekohl MY_FORK_URL`

Now my remotes look like this (I also do some rewriting to pull over HTTPS):
```console
$ git remote -v
ekohl	https://github.com/ekohl/foreman (fetch)
ekohl	git@github.com:ekohl/foreman (push)
upstream	https://github.com/theforeman/foreman (fetch)
upstream	git@github.com:theforeman/foreman (push)
```

That brings me to naming: isn't this really more `clone_from_fork`?

Now a deeper question is: why do you want this?

ekohl:
> ```shell
> 2022-08-04 20:02:57 [ERROR ] [configure] /Stage[main]/Katello_devel::Install/Katello_devel::Git_repo[foreman]/Vcsrepo[/home/vagrant/foreman]: Could not evaluate: Execution of 'git fetch chris1984' returned 128: fatal: could not read Password for 'https://chris1984@github.com': No such device or address
> ```

This typically happens if you're trying to clone/pull from a GitHub repository URL that doesn't exist. For example, if you haven't cloned something.

chris1984 (approved):
ACK

After forking Foreman and REX it works great:

```bash
PLAY RECAP *********************************************************************
centos82-katello-devel     : ok=58   changed=28   unreachable=0    failed=0    skipped=27   rescued=0    ignored=0

[vagrant@centos82-katello-devel foreman]$ git remote -v
chris1984	https://chris1984@github.com/chris1984/foreman.git (fetch)
chris1984	https://chris1984@github.com/chris1984/foreman.git (push)
upstream	https://github.com/theforeman/foreman.git (fetch)
upstream	https://github.com/theforeman/foreman.git (push)
```

wbclark:
Great question @ekohl 

My typical workflow is similar to yours but reversed

`git clone -o wbclark FORK_URL`
`git remote add upstream UPSTREAM_URL`

I'd like the option to setup the devel environment this way as well, to mirror what has become a comfortable setup for me.

Looking forward, it would also be nice to specify additional forks such as a list of team members (for doing code reviews) or downstream and wbclark-downstream forks.

In the former case it would be as simple as passing an additional Array of Github usernames, but in the latter case there would be additional complexity of needing to specify both the remote name and complete URL.

Then there is the question of whether additional remotes should be for the Katello repository only, or whether they could be passed for foreman, remote execution, etc.

My goal was to implement this first and then come back and revisit the hard parts.

I think the `clone_from_fork` name is better and I'm happy to make that change.

ekohl:
> Looking forward, it would also be nice to specify additional forks such as a list of team members (for doing code reviews) or downstream and wbclark-downstream forks.

For upstream work I'd recommend looking at [hub](https://hub.github.com/), in particular https://hub.github.com/#maintainer. Looks like it's not in EPEL (https://src.fedoraproject.org/rpms/golang-github-hub) but since it's go there's just a single binary. You could also consider requesting an EPEL branch for it.

I can't imagine working with GitHub without hub.

ekohl (commented):
Thinking more about this, I think this is a bad workflow. IMHO the default branch (`master`, `develop`) should always track upstream. A simple `git pull` in `master` now follows your fork, no longer the project's master. We've seen with new people that they end up confused because their master is not the most recent and rebases end up based on very old commits. You can also see what @chris1984 ran into: if you haven't forked it, you can't clone it. From the password prompt it's not obvious what went wrong.

I'm going to say :-1: on cloning from forks because it can only create more confusion.

If the problem is that pushing often happens to the wrong remotes, perhaps that's an easier problem to solve?

I have an alias set up. We can run the equivalent of `git config --global alias.ph '!git push ${katello_devel::fork_remote_name_real} HEAD --set-upstream'`. That way the workflow becomes:

```sh
git fetch upstream
git checkout -b my-branch upstream/master
# do work
git ph
hub pull-request
```

At least, that's roughly what I always do.

Now I also read `man git switch` and that mentions an option `checkout.defaultRemote`. Perhaps also an interesting one to consider.

But overall I struggle a bit with the balance of "get me a basic working environment" and "allow me to fully tailor my environment". For the last part I use https://github.com/RichiH/vcsh btw.

wbclark:
Closing this based on feedback from @ekohl .

For my personal use case, I have a role in forklift that can provide what I need now.

Thanks all for the thoughtful discussion and recommendations.

PR #284, opened 2022-08-23
Add an optional parameter foreman_custom_remotes

wbclark:


wbclark:
> IMHO this really needs more explanation. I'd really like to avoid completely customizing everything using Puppet. As I said in [#283 (comment)](https://github.com/theforeman/puppet-katello_devel/pull/283#issuecomment-1211773503), if you need to review PRs, use hub.

Sure, so reviewing is one of the use cases I envisioned but it is more secondary.

I would really like to be able to set up access to other remotes in a development environment. That way I could cherry-pick an upstream patch into a downstream branch, build the source, build an RPM, etc.

And others could do the same, without requiring a lot of setup and configuration on their workstation --> just copy some yaml into their forklift box definitions and vagrant up, which is already a familiar process.

wbclark:
Closing this, since https://github.com/theforeman/puppet-katello_devel/commit/62324573a95734c7a72c54b4d85f5c4722679499 allows the user to manage the foreman source repository independently of this module

PR #285, opened 2022-09-09
Refs #33956 - proxy assets and webpack to Rails in devel

evgeni:
In https://github.com/theforeman/puppet-foreman/commit/215db6539dbfdbd2896946537f1e520fac7f63ba we made a change that `/webpack` and `/assets` are served by Apache by default. This is correct for production setups, but breaks development ones, as here `/assets` is actually served by Rails.

Explicitly configure `foreman::apache` to not include these in the "no_proxy" list.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -8,6 +8,7 @@ class katello_devel::apache {
     ssl_ca        => $certs::apache::ca_cert,
     ssl_chain     => $certs::apache::apache_ca_cert,
     proxy_backend => "http://localhost:${katello_devel::rails_port}/",
+    proxy_assets  => true,
   }
 
   # required by configuration in katello/katello-apache-ssl.conf


evgeni:
@ekohl I wonder if we should make this a flag in `foreman::config::apache` `Boolean proxy_assets` or something?

evgeni:
Also, technically, we can still include `/webpack` in the no_proxy list, as webpack runs on a different port in development, but `/assets` is straight Rails.

ekohl:
> @ekohl I wonder if we should make this a flag in `foreman::config::apache` `Boolean proxy_assets` or something?

I'd be happy with that.

evgeni:
> > @ekohl I wonder if we should make this a flag in `foreman::config::apache` `Boolean proxy_assets` or something?
> 
> I'd be happy with that.

https://github.com/theforeman/puppet-foreman/pull/1085

evgeni:
@ekohl this is now using the new interface and ready to merge

ekohl (approved):
Thanks!

PR #286, opened 2022-09-09
Use NodeJS 14

ekohl:


evgeni:
According to @mariaaga it should still work with 12, so I am not sure we should flip the default just yet?

MariaAga:
Everyone should use node 14 technicality as node 12 is eol for 4 months now :woman_shrugging: 
https://endoflife.date/nodejs

ekohl:
Now I'm wondering what we use in our build environment.

Looks like in our CI we still use Node 12:
https://github.com/theforeman/foreman-infra/blob/900c13c60bd1184ef4a5773e99ce13239f0ccb1c/puppet/modules/slave/manifests/unittests.pp#L57-L61
Debian also builds with 12:
https://github.com/theforeman/foreman-infra/blob/900c13c60bd1184ef4a5773e99ce13239f0ccb1c/puppet/modules/slave/files/pbuilder_F66-add-nodesource-repos#L4
And RPMs too if we still use the flatten-el8 script for our build roots:
https://github.com/theforeman/foreman-infra/blob/900c13c60bd1184ef4a5773e99ce13239f0ccb1c/koji/flatten-el8/rhel8-split.sh#L71

evgeni:
Correct, and we talked in the past about updating these (https://community.theforeman.org/t/javascript-infrastructure-update-plan/28614) but didn't go that far yet.

ekohl:
Thanks for reminding me of that. In https://community.theforeman.org/t/path-to-ruby-3-0-3-1-el9-and-ubuntu-22-04/30295/4 I've also considered that aspect. On EL9 the default NodeJS version is 16 (with 18 in a module) so that's another incentive to update.

wbclark:
to what extent is this blocked by the infrastructure tasks previously mentioned?

I mention this because a colleague faced an issue with the katello dev environment and it turned out the reason was that they were still using NodeJS 12.

ekohl:
We're also using NodeJS in our packaging, so if it's broken in your development setup, it'll also be broken in our building. https://community.theforeman.org/t/resurrection-of-the-client-side-infrastructure-upgrade-effort/32340 was just posted, but it's going to be a massive effort.

For now  I consider this a draft, until we can actually move forward.

m-bucher:
Looks like `npm` (or rather some node-packages) complain about this, when doing an `npm install` in forklift:
```
npm ERR! code ENOTSUP
npm ERR! notsup Unsupported engine for expect@29.6.4: wanted: {"node":"^14.15.0 || ^16.10.0 || >=18.0.0"} (current: {"node":"12.22.12","npm":"6.14.18"})
npm ERR! notsup Not compatible with your version of node/npm: expect@29.6.4
npm ERR! notsup Not compatible with your version of node/npm: expect@29.6.4
npm ERR! notsup Required: {"node":"^14.15.0 || ^16.10.0 || >=18.0.0"}
npm ERR! notsup Actual:   {"npm":"6.14.18","node":"12.22.12"}
```
Therefore, I'd welcome this moving forward, at least for the katello-devel boxes :grin: 

I must admit it does not seem to be critical, since the command apparently finishes successful.

PR #288, opened 2022-11-03
Allow Katello source repository to be unmanaged

wbclark:
This builds on #287 and allows the Katello repository specifically to be unmanaged

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -6,6 +6,7 @@ class katello_devel::config (
   String $group = $katello_devel::group,
   Array[Variant[String,Hash]] $extra_plugins = $katello_devel::extra_plugins,
   String $katello_scm_revision = $katello_devel::katello_scm_revision,
+  Boolean $katello_manage_repo = $katello_devel::katello_manage_repo,
 ) {
   file { "${foreman_dir}/.env":
     ensure  => file,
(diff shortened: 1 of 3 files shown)


manifests/config.pp
```diff
@@ -33,6 +34,7 @@
   katello_devel::plugin { 'katello/katello':
     settings_template => 'katello_devel/katello.yaml.erb',
     scm_revision      => $katello_scm_revision,
+    manage_repo       => $katello_manage_repo,
```
ehelms, line 37: Secondary thought -- move this katello declaration out of here and configure `extra_plugins` here by default for Katello? https://github.com/theforeman/forklift/blob/master/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml#L6

Not a blocking request, just a thought on normalizing plugins a bit more in the module.

wbclark: My first thought is that would look like moving the `settings_template` and `scm_revision` parameters into the hash as optional keys, but that breaks the `katello_devel` class interface which already has `katello_scm_revision` as a top level param.

Another thing to consider is the fact that katello is much more tightly coupled than a generic plugin, and requires additional services to be deployed, so should it be treated as no different from other plugins?

If we only deploy pulp, candlepin, etc. when deploying the katello plugin, this module then starts to look like it should be called `puppet-foreman_devel` but would it actually have everything needed for someone who is developing foreman without katello?

Thinking about it more, I might lean the other way entirely if I were building this module from scratch, and just split the katello plugin off into its own class that doesn't involve any `katello_devel::plugin` resource at all.

ehelms (approved):

PR #291, opened 2023-08-12
Remove katello-agent

wbclark:


wbclark:
> Should we also ensure it's absent for those who update their devel box?

Devel boxes are cheap and get rebuilt fairly often. I think it's probably fine if we ask people to rebuild in this case.

Just to be sure, I asked the team about this during standup today, and consensus was that it's fine not to worry about it.

ekohl (approved):

PR #294, opened 2023-09-28
Update gems to support testing Puppet 8

ekohl:


ehelms (approved):

ekohl:
CI didn't ran because it was auto disable (due to the cron resource and no changes). I'll make it a required step so we don't forget again, but first https://github.com/theforeman/puppet-katello_devel/pull/295 needs to pass.

ehelms:
> CI didn't ran because it was auto disable (due to the cron resource and no changes). I'll make it a required step so we don't forget again, but first #295 needs to pass.

ARG, I hate that feature of Github Actions.

ekohl:
You're not alone. I fear we may need to copy the workflows to get both cron and PR testing. On the other hand, I don't think we really look at the cron that often so does it even make sense to keep it?

ehelms:
> You're not alone. I fear we may need to copy the workflows to get both cron and PR testing. On the other hand, I don't think we really look at the cron that often so does it even make sense to keep it?

There's a cron job? Guess that answers your question.

PR #300, opened 2024-06-05
Support EL9 and deprecate CentOS-8

wbclark:


wbclark:
I used https://github.com/voxpupuli/beaker-hostgenerator/pull/353 while running the acceptance test locally, so it should work with the newer CentOS 9 Stream images as well.

spec/setup_acceptance_node.pp
```diff
@@ -15,14 +15,17 @@
 
 # Necessary for PostgreSQL EVR extension
 yumrepo { "pulpcore":
-  baseurl  => "http://yum.theforeman.org/pulpcore/3.14/el\$releasever/x86_64/",
+  baseurl  => "http://yum.theforeman.org/pulpcore/nightly/el\$releasever/x86_64/",
   descr    => "Pulpcore",
   enabled  => true,
-  gpgcheck => true,
-  gpgkey   => "https://yum.theforeman.org/pulpcore/3.14/GPG-RPM-KEY-pulpcore",
```
wbclark, line 22: Hey @zjhuntin , not sure if this was intentional or not, but there's no GPG key in the nightly RPM builds for pulpcore.

It doesn't really matter for the purposes of this PR, since this is for CI and we'd be fetching the GPG key from the same location as the RPMs, using it once, and then tearing it down, but I wanted to give you as a heads up just in case it matters more to somebody somewhere else. ;-)

metadata.json
```diff
@@ -51,13 +51,14 @@
     {
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
-	      "8"
+	      "8",
+	      "9"
```
ekohl, line 55: Looks like this contains tabs. Mind fixing it while you're at it? Same below for CentOS

ekohl (approved):
Thanks. Perhaps add AlmaLinux 8 & 9 as well (obsoleting https://github.com/theforeman/puppet-katello_devel/pull/301) to make sure we keep an EL 8 devel box supported for now?

wbclark:
> Thanks. Perhaps add AlmaLinux 8 & 9 as well (obsoleting #301) to make sure we keep an EL 8 devel box supported for now?

Done, thanks.

(The acceptance test failed for AlmaLinux due to bundle install timeout and oom... let's see if that's an issue on CI infra...)

wbclark:
:green_apple:

PR #301, opened 2024-06-06
Add AlmaLinux 8 & 9 support

archanaserver:
Added AlmaLinux 8 and 9 to the supported operating systems in metadata.json.

These changes ensure that we start testing with AlmaLinux 8 and 9 instead of CentOS 8 Stream. This update is part of our transition to better support AlmaLinux in our Puppet modules.

See, for details: https://github.com/theforeman/foreman-infra/issues/2087

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -59,6 +59,13 @@
       "operatingsystemrelease": [
 	      "8"
       ]
+    },
+    {
+      "operatingsystem": "AlmaLinux",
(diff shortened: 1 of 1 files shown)


ekohl:
Looks like the repository setup on EL 9 is broken. There previously was no EL 9 test, so it may also be an issue with CentOS 9.

ekohl:
Implemented in #300

PR #302, opened 2024-06-10
Enable CRB on EL9 always

wbclark:


evgeni:
That's needed for libvirt-devel being installable?

wbclark:
> That's needed for libvirt-devel being installable?

Yes, exactly

evgeni:
CI hates you (probably missing fixtures entry for yumrepo_core?), but otherwise ACK

wbclark:
> CI hates you (probably missing fixtures entry for yumrepo_core?), but otherwise ACK

Yeah, trying to figure out why it was fine when we did it in setup_acceptance_node.pp but not here D:

evgeni:
The failure is in the unit tests, which runs on the gem puppet that has no core modules. Acceptance runs AIO that has them.

metadata.json
```diff
@@ -28,6 +28,10 @@
       "name": "puppetlabs-vcsrepo",
       "version_requirement": ">= 1.0.0"
     },
+    {
+      "name": "puppetlabs-yumrepo_core",
```
evgeni, line 32: That's not needed, IMHO

wbclark: ah, I see the explanation about gem puppet vs. AIO... I'll remove it in a moment

Thanks!

evgeni (approved):

PR #303, opened 2024-06-12
Use RHEL flavor of CRB on RHEL

adamruzicka:
This depends on a puppet/rhsm module. Considering this is a bit of an edge case, I put it in as an optional dependency.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -2,9 +2,16 @@
 # @api private
 class katello_devel::install {
   if $facts['os']['release']['major'] == '9' {
-    yumrepo { 'crb':
-      enabled => true,
-      before  => Package['libvirt-devel'],
+    if $facts['os']['name'] == 'RedHat' {
+      rh_repo { "codeready-builder-for-rhel-${facts['os']['release']['major']}-${facts['os']['architecture']}-rpms":
+        ensure => present,
+        before => Package['libvirt-devel'],
+      }
+    } else {
+      yumrepo { 'crb':
+        enabled => true,
+        before  => Package['libvirt-devel'],
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


manifests/install.pp
```diff
@@ -2,9 +2,16 @@
 # @api private
 class katello_devel::install {
   if $facts['os']['release']['major'] == '9' {
-    yumrepo { 'crb':
-      enabled => true,
-      before  => Package['libvirt-devel'],
+    if $facts['os']['distro']['id'] == 'RedHatEnterprise' {
```
ekohl, line 5: You can check for the OS **name**, where `RedHat` is RHEL (as opposed to the OS **family** fact)
```suggestion
    if $facts['os']['name'] == 'RedHat' {
```

manifests/install.pp
```diff
@@ -2,9 +2,16 @@
...
+      rh_repo { "codeready-builder-for-rhel-9-${facts['os']['architecture']}-rpms":
+        ensure => present,
+        before  => Package['libvirt-devel'],
```
ekohl, line 8:
```suggestion
        before => Package['libvirt-devel'],
```

ekohl (changes requested):

manifests/install.pp
```diff
@@ -2,9 +2,16 @@
 # @api private
 class katello_devel::install {
   if $facts['os']['release']['major'] == '9' {
-    yumrepo { 'crb':
-      enabled => true,
-      before  => Package['libvirt-devel'],
+    if $facts['os']['name'] == 'RedHat' {
+      rh_repo { "codeready-builder-for-rhel-9-${facts['os']['architecture']}-rpms":
```
ekohl, line 6: I saw that in CentOS Stream 10 also has a CRB repo, so I think we can assume RHEL 10 has the same structure (even if line 4 prevents it from being executed).  Should we prepare for that?
```suggestion
      rh_repo { "codeready-builder-for-rhel-${facts['os']['release']['major']}-${facts['os']['architecture']}-rpms":
```

ekohl (approved):

ekohl (changes requested):
You need to add the module to `.fixtures.yml` for CI to be happy

ekohl (approved):

PR #307, opened 2024-09-19
Allow specifying extra gemfiles for a plugin

ehelms:
This aims to provide a targeted solution to extra gemfiles rather than a catch all solution.

This should prevent the need for https://github.com/theforeman/foreman_virt_who_configure/pull/201 and to keep the solution in place of having a `test.rb` in Katello.


manifests/plugin.pp
```diff
@@ -9,10 +9,14 @@
...
 # @param manage_repo
 #   Set to false if the plugin source repository is managed externally.
+#
+# @param extra_gemfiles
+#   Additional gemfiles a plugin needs added
 define katello_devel::plugin (
   Optional[String] $settings_template = undef,
   Optional[String] $scm_revision = undef,
   Boolean $manage_repo = true,
+  Array[String] $extra_gemfiles = [],
```
chris1984, line 19: It has been a long time since I looked at puppet, here we are setting an empty array, then in https://github.com/theforeman/puppet-katello_devel/pull/307/files#diff-b55cd2d4fee4ad88a9e21316ff2015c5579e8ec9bef7bc0fe5253262905fa702R51 we are hard coding it, is that so Katello always has it and we can override it if needed?

ehelms: This is a `define` which is meant to be a re-usable chunk of code, so we assume this is empty by default and plugin definitions can pass in values if they need.

ehelms:
> We can also consider the use of groups (like `ci`) and configure Bundler to disable them in dev. You can also mark a group as optional and enable it in our CI. Though I have a vague recollection Bundler always evaluates all dependencies so it may break. Alternatives are to guard it by env vars in the repo

I don't think that helps us here.

chris1984 (approved):
Applied the patch and it works now that rh_cloud and virt-who are green:

```bash
[root@toledodevel ~]# foreman-installer -v --no-colors   --scenario katello-devel --foreman-proxy-content-enable-ostree=true --katello-devel-modulestream-nodejs=18 --katello-devel-admin-password changeme  --katello-devel-extra-plugins theforeman/foreman_rh_cloud --katello-devel-extra-plugins theforeman/foreman_virt_who_configure
2024-09-24 10:16:27 [NOTICE] [root] Loading installer configuration. This will take some time.
2024-09-24 10:16:31 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
2024-09-24 10:16:31 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
2024-09-24 10:16:33 [NOTICE] [checks] System checks passed
2024-09-24 10:16:40 [NOTICE] [configure] Starting system configuration.
2024-09-24 10:16:54 [NOTICE] [configure] 250 configuration steps out of 1338 steps complete.
2024-09-24 10:16:56 [NOTICE] [configure] 500 configuration steps out of 1339 steps complete.
2024-09-24 10:16:59 [NOTICE] [configure] 750 configuration steps out of 1342 steps complete.
2024-09-24 10:17:03 [NOTICE] [configure] 1000 configuration steps out of 1347 steps complete.
2024-09-24 10:36:04 [NOTICE] [configure] 1250 configuration steps out of 1347 steps complete.
2024-09-24 10:36:28 [NOTICE] [configure] System configuration has finished.
  Success!
  * To run the Katello dev server log in using SSH
  * Run `cd foreman && bundle exec foreman start`
  * The server is running at https://toledodevel.satellite.lab.eng.rdu2.redhat.com
  * On Firefox you need to accept the certificate at https://toledodevel.satellite.lab.eng.rdu2.redhat.com
:3808
      Initial credentials are admin / changeme
  * Foreman Proxy is running at https://toledodevel.satellite.lab.eng.rdu2.redhat.com:9090

The full log is at /var/log/foreman-installer/katello-devel.log
```

@ehelms can this get in so we can spin up new devel boxes?

PR #311, opened 2025-04-17
Drop rake webpack:compile

adamruzicka:
For some reason this seems to depend on db being migrated, currently the webpack compile step fails like this https://gist.github.com/lumarel/3d04de883cc14deb7f5cff12b0ea2e30 .

Afaik it shouldn't be necessary anymore, if foreman is started with the provided procfile, webpack should be watching the directory and compiling assets on the fly.

An alternative fix would be to change the ordering.


ekohl (approved):
I don't have a strong opinion on this

PR #313, opened 2025-07-08
feat(db): require password for local IP connections

vkrizan:
Requires password authentication to pg for local IP connections in order to support Foreign Data Wrapper connections to the same instance.

Please note, that the unix socket connections are unaffected by this change.

Having a FDW to `localhost` with full `trust` results in:
```
ERROR:  password is required
DETAIL:  Non-superuser cannot connect if the server does not request a password.
HINT:  Target server's authentication method must be changed or password_required=false set in the user mapping attributes. 
```

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -33,12 +33,30 @@ class katello_devel::database {
     auth_method => 'trust',
   }
 
+  postgresql::server::pg_hba_rule { 'host IPV4 inventory_db':
+    type        => 'host',
+    database    => 'inventory_db',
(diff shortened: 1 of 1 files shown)


vkrizan:
@ehelms looks like this breaks something? Should I just add `inventory_db` and leave the rest untouched?

vkrizan:
> Will the same be required for advisor?
@ehelms This will be shared for everything that needs to grab data from Inventory db, Advisor included. The trick here is to require password for the database/user it is connecting to (via foreign table).

PR #324, opened 2026-08-20
Switch from puppet to OpenVox 8

bastelfreak:
contains https://github.com/theforeman/puppet-katello_devel/pull/323


bastelfreak:
requires https://github.com/theforeman/foreman-installer-modulesync/pull/181

